PR #12, opened 2016-05-19
Add attributes to Resources.

saeta:
Documentation is represented as an attribute. This commit adds in the forgotten
field to the courier models.


yifan-coursera:
LGTM.

PR #19, opened 2016-05-23
Configure bintray publishing for sbt plugin.

saeta:
This commit adds all the configuration & publishes the sbt-naptime plugin
to bintray. It has been marked for inclusion in the standard sbt plugins
bintray distribution, but is not available yet. If you would like to depend
on it presently, add the following to `project/plugins.sbt`:

```
resolvers += Resolver.bintrayIvyRepo("coursera", "sbt-plugins")
```

This commit also fixes a publishing problem in the maven central publishing workflow.


yifan-coursera:
LGTM.

PR #20, opened 2016-05-23
Remove unused bintray repository.

saeta:
The courier sbt plugin is now mirrored in the standard sbt plugin bintray
distribution, so we can remove our own resolver.

Note: in this commit, we comment out the travis-ci caching, in order to
properly test that this change is safe.


saeta:
Based on https://travis-ci.org/coursera/naptime/builds/132420142, it looks like everything's building properly, so I'm goign to revert the commenting out of the build cache in https://github.com/coursera/naptime/pull/20/commits/d08de066cd597ba729369d39202b1bce05371dbb

PR #22, opened 2016-05-24
Add a NoRestrictions authenticator.

saeta:
In order to provide a default authentication in a number of places in the codebase,
we add a NoRestrictions authenticator that allows all requests to pass.


yifan-coursera:
LGTM.

PR #24, opened 2016-05-25
More flexible header parsing.

saeta:
In order to support more sophisticated header parsing, I am adding the ability
to customize the response code associated with the Error parsing response. In
particular, if the header is malformed for whatever reason, it makes more sense
to return a BAD_REQUEST than a FORBIDDEN response code.

Additionally, I am removing the unnecessary NoRestrictions parser.


naptime/src/main/scala/org/coursera/naptime/access/StructuredAccessControl.scala
```diff
@@ -44,8 +45,231 @@ case class StructuredAccessControl[A](
...
     }
   }
 }
+
+object StructuredAccessControl extends StructuredAccessControlCombinators {
+  private[access] val missingResponse =
+    Left(NaptimeActionException(Status.UNAUTHORIZED, Some("auth.perms"), Some("Missing authentication")))
+}
+
+trait StructuredAccessControlCombinators {
```
josh-newman: I think these should be structured the same way as the authenticator combiners to keep everything consistent and maintainable. A few changes:
* Move these to a new `combiner` subpackage.
* Rename helpers to `anyOf`, and `eitherOf`.
* Move to separate files: AnyOf.scala, And.scala, EitherOf.scala.

(Also I'm not sure how I originally chose "combiner" vs. "combinator", but we should stick with one.)

naptime/src/main/scala/org/coursera/naptime/access/StructuredAccessControl.scala
```diff
@@ -44,8 +45,231 @@ case class StructuredAccessControl[A](
...
+   * left one.
+   */
+  def either[A, B](
+      left: StructuredAccessControl[A],
+      right: StructuredAccessControl[B]): StructuredAccessControl[Either[A, B]] = {
+    val authn = new Authenticator[Either[A, B]] {
+      override def maybeAuthenticate(
+          requestHeader: RequestHeader)
+          (implicit ec: ExecutionContext): Future[Option[Either[NaptimeActionException, Either[A, B]]]] = {
+        val lhsF = left.authenticator.maybeAuthenticate(requestHeader)
```
josh-newman: Wrap these in `Futures.safelyCall`.

naptime/src/main/scala/org/coursera/naptime/access/StructuredAccessControl.scala
```diff
@@ -44,8 +45,231 @@ case class StructuredAccessControl[A](
...
+   *     StructuredAccessControl.either(Auth.superuser, Auth.admin)).get(ctx => ???)
+   * }}}
+   *
+   * If there is only one parsing error, it is ignored. If there are two, the right one masks the
+   * left one.
+   */
+  def either[A, B](
+      left: StructuredAccessControl[A],
+      right: StructuredAccessControl[B]): StructuredAccessControl[Either[A, B]] = {
+    val authn = new Authenticator[Either[A, B]] {
```
josh-newman: Call this `authenticator`, to stick with our style of avoiding abbreviations.

naptime/src/main/scala/org/coursera/naptime/access/StructuredAccessControl.scala
```diff
@@ -44,8 +45,231 @@ case class StructuredAccessControl[A](
...
+   * A left-biased combinator that allows the request if either the left or the right access
+   * control allow it.
+   *
+   * Example usage:
+   * {{{
+   *   Nap.auth(
+   *     StructuredAccessControl.either(Auth.superuser, Auth.admin)).get(ctx => ???)
+   * }}}
+   *
+   * If there is only one parsing error, it is ignored. If there are two, the right one masks the
```
josh-newman: Parsing errors are specific; authenticators can throw arbitrary exceptions. I assume all errors are handled the same way?

Why do errors pick the right if success is biased to the left?

saeta: Removed `parsing` as it is imprecise. The reason why is because when the left has an error, we fallback to whatever happens on the `Right`.

PR #34, opened 2016-06-01
Extract and invoke authorizer in `RestActionTester`

josh-newman:
@saeta, what do you think of this as an alternative to #33? This isn't complete but we only use `StructuredAccessControl`s right now, I think, so it should work temporarily. Longer term, I want to think more about ideal testing patterns for our actions, but that discussion can be more leisurely after this is fixed.


saeta:
But we do use the combiners heavily, which don't extend from
StructuredAccessControl

saeta:
Per discussion in person, I'm going to close this one @josh-newman until we have the longer conversation regarding access control. (Perhaps I'll start a (public) doc that'll include some of my other thoughts for access control, including field-level access control, and element-level access control too...)

PR #35, opened 2016-06-01
Treat None.type specially

yifan-coursera:
Previously, `SchemaInference.inferSchema` would break on schemas that included `None.type`. This change handles `None.type` correctly. 

When evaluating the following code:
```
import org.coursera.naptime.courier.SchemaInference
import scala.reflect.runtime.universe

case class WithNone(nothing: None.type)
SchemaInference.inferSchema(universe.typeOf[WithNone])
```
Before:
```
java.lang.ClassCastException: scala.reflect.internal.Types$UniqueSingleType cannot be cast to scala.reflect.api.Types$TypeRefApi
  at org.coursera.naptime.courier.ScalaClassTraverser.org$coursera$naptime$courier$ScalaClassTraverser$$inferSchema(SchemaInference.scala:118)
```

After:
```
res0: play.api.libs.json.JsObject = {"name":"WithNone","type":"record","fields":[{"name":"nothing","type":"null"}],"namespace":"$line5"}
```

Additionally, we change the schema for `Some[A]` to be identical to the schema of `A`.


saeta:
Thanks for this @yifan-coursera !! I'm merging this now. When do you need me to cut a new release of naptime?

PR #38, opened 2016-06-02
HeaderAccessControl.either should be right-biased

saeta:
This follows the conventions in other programming languages (e.g. Haskell).


danchia:
LGTM

PR #40, opened 2016-06-06
Change asymmetric type computation log message from error to warning

yifan-coursera:
Since the these methods still return valid values even if errors occurred, errors should be logged as warnings rather than errors.


saeta:
`s/warning/warn/g`?

PR #41, opened 2016-06-07
Use type inference for non-Courier types

yifan-coursera:
When encountering a non-Courier type, run SchemaInference on the type to infer the schema if possible. If errors result, fall back to returning no schema.


saeta:
This looks good to me.

Just to confirm, we need to extend the `Resource.courier` schema to include an extra field for the un-adulterated value type, correct?

yifan-coursera:
Yup. I think we should rename `bodyType` to `mergedType` and have `bodyType` be the unadulterated type, but that involves an interface change.

saeta:
Let's take the breaking change sooner rather than later. I'll send a follow-on PR for this shortly.

PR #43, opened 2016-06-08
RestEngine2's: Carefully test resources

saeta:
In order to ensure a smooth migration to the 2nd-generation engines,
I've fleshed out a more sophisticated test suite that ensures that
the new engines behave appropriately.


saeta:
This should be merged only after https://github.com/coursera/naptime/pull/42 lands (it builds on top of it).

PR #44, opened 2016-06-08
Add a linter and begin fixing lint warnings

saeta:
Right now, most things are warning. Over time, we'll turn
lints into errors, and we'll hook it into our CI jobs as
well.


saeta:
PTAL @yifan-coursera

PR #45, opened 2016-06-08
Augment Resource.courier to include mergedType

saeta:
As part of the asymmetric models approach we've adopted, we compute a
merged type that is composed of the union of the fields between the
key and the value. Sometimes, however, it's important to see the
unadulterated value type. For this, we now expose both the valueType
and the mergedType. Previously, the exposed bodyType was the mergedType,
and the valueType was not exposed at all. Hopefully with this rename,
we've also settled on good names for these types.


saeta:
PTAL @yifan-coursera 

This is the follow-on to the discussion in: https://github.com/coursera/naptime/pull/41#issuecomment-224445304

PR #47, opened 2016-06-09
Add tests for ETag construction.

saeta:
Add some tests to verify correct ETag behavior for the
Engine2 implementations.


saeta:
PTAL @yifan-coursera

PR #48, opened 2016-06-12
Add If-None-Match tests for Engine2's

saeta:
Test the engine2 implementations to ensure they perform correctly
in the presence of If-None-Match headers and correctly return
HTTP 304 NOT MODIFIED responses with empty bodies.


saeta:
PTAL @yifan-coursera

PR #50, opened 2016-06-15
Use Engine2's by default.

saeta:
Engine2's have been tested in production @ Coursera, and
they seem to be behaving. Therefore, we're going to roll
forward and use Engine2's by default. (You can roll back
by importing the old PlayJson engines.)


saeta:
PTAL @zhaojunz

zhaojunz:
not sure that I am qualified for review. lgtm to me. maybe @josh-newman could take a look?

PR #53, opened 2016-06-24
Define strong and weak ETags separately

josh-newman:
@saeta

### DIFF
--- a/naptime/src/main/scala/org/coursera/naptime/ETag.scala
+++ b/naptime/src/main/scala/org/coursera/naptime/ETag.scala
@@ -19,50 +19,24 @@ package org.coursera.naptime
 import org.coursera.common.stringkey.StringKeyFormat
 import org.coursera.common.stringkey.StringKeyFormat.Implicits.OrFormat
 
-sealed trait ETag
+/**
+ * Note: Naptime always uses weak validators because the framework does not guarantee byte-equality
+ * between responses.
+ *
+ * @param weakValidator tag without surrounding `"`s; these are added by Naptime
+ */
+case class ETag(weakValidator: String)
 
 object ETag {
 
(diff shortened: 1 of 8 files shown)


naptime/src/main/scala/org/coursera/naptime/ETag.scala
```diff
@@ -0,0 +1,68 @@
...
+object ETag {
+
+  implicit val stringKeyFormat: StringKeyFormat[ETag] = {
+    StringKeyFormat.unimplementedFormat[ETag]
+      .orFormat[Strong]
+      .orFormat[Weak]
+  }
+
+  /**
+   * TODO(josh): Should Naptime support this at all? Play JSON serialization doesn't allow
```
saeta: Could you actually just switch everything to use the weak etags? I believe that is the most semantically correct thing to do.

naptime/src/main/scala/org/coursera/naptime/ETag.scala
```diff
@@ -0,0 +1,42 @@
...
+import org.coursera.common.stringkey.StringKeyFormat
+import org.coursera.common.stringkey.StringKeyFormat.Implicits.OrFormat
+
+/**
+ * Note: Naptime always uses weak validators because the framework does not guarantee byte-equality
+ * between responses.
+ *
+ * @param weakValidator tag without surrounding `"`s; these are added by Naptime
+ */
+case class ETag(weakValidator: String)
```
saeta: Should we require that `weakValidator` does not contain a `"`?

saeta:
@josh-newman looks like there are some important test failures. Could you update them? After that, then LGTM from me.

josh-newman:
@saeta, PTAL.

saeta:
@josh-newman LGTM.

PR #59, opened 2016-06-29
Add strong ETags for backwards-compatibility

josh-newman:
@saeta 

I've marked them as deprecated already because they're not actually correct to use and we should remove them as soon as we fix the few existing APIs.


saeta:
LGTM.

PR #60, opened 2016-07-11
Add bad gateway as error option

davidswinegar:
Add a 502 error (bad gateway) as standard naptime error option.
Also fix an incorrect comment.

@josh-newman @saeta


saeta:
@davidswinegar can you describe the reason for needing a 502 response?

davidswinegar:
@saeta I'm currently throwing a `NaptimeActionException` instead of using this shortcut. I need it to indicate that errors are due to a upstream service configuration error, but are not actionable for non-privileged clients.

PR #61, opened 2016-07-19
Add `successfulOf` combinator to `HeaderAccessControl`

josh-newman:
@saeta


saeta:
LGTM, except could you remove the version bump?

josh-newman:
Reverted. Why no version bump? Do you prefer to do that right before publishing?

saeta:
Yeah, exactly. It's also where I've been putting the "release notes / changes" for now.

PR #62, opened 2016-07-19
Introduce HTTP-independent `AccessControl`

josh-newman:
@saeta 

Also included:
* I renamed access control's `check` to describe its purpose better and added various Scaladoc.
* I fixed a small bug where `HeaderAccessControl.and` arity 3 ignored the last access control's failure.

### DIFF
--- a/naptime/src/main/scala/org/coursera/naptime/access/AccessControl.scala
+++ b/naptime/src/main/scala/org/coursera/naptime/access/AccessControl.scala
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2016 Coursera Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.coursera.naptime.access
+
+import org.coursera.naptime.NaptimeActionException
+
+/**
+ * Control access to an API using Naptime's DSL.
+ *
+ * Implementation note: This access control abstraction is independent of whether requests arrive
+ * via HTTP or other means; until Naptime supports those other means, [[HeaderAccessControl]] is
+ * the only access control type. When that's supported, this trait will likely have more methods
+ * that make it more directly useful.
+ *
+ * @tparam A structured authentication result which is available to the API implementation after
+ *           access control is successful
+ */
+trait AccessControl[A] {
+
+  /**
+   * Used for exercising access control configurations in resource tests.
+   */
+  private[naptime] def check(authInfo: A): Either[NaptimeActionException, A]
+
+}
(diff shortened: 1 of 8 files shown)


naptime/src/main/scala/org/coursera/naptime/access/AccessControl.scala
```diff
@@ -0,0 +1,47 @@
...
+
+  final def runAndCheck(request: R)(implicit ec: ExecutionContext): Future[A] =
+    run(request).map(_.left.map(throw _).merge)
+
+  /**
+   * Used for exercising access control configurations in resource tests. Simulates the result of
+   * the authentication process (since that often involves data fetching, and should be unit tested
+   * separately from resources).
+   */
+  private[naptime] def simulateAuthentication(authentication: A): Either[NaptimeActionException, A]
```
saeta: Shouldn't this be `simulateAuthorization`?

naptime/src/main/scala/org/coursera/naptime/access/AccessControl.scala
```diff
@@ -0,0 +1,47 @@
...
+import org.coursera.naptime.NaptimeActionException
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+
+/**
+ * Control access to a resource method. Arbitrary, asynchronous logic may be executed, but
+ * successful access grant must produce an authentication object of type [[A]].
+ *
+ * @tparam R request data (for example, for Play HTTP requests, this is
```
saeta: What's the advantage of making this a type parameter at this point?

saeta:
Hey, I guess I'm confused as to the larger context. What's the purpose of the new `AccessControl` trait?

naptime/src/main/scala/org/coursera/naptime/access/AccessControl.scala
josh-newman: @saeta: With just this PR, it's not useful, because it's not used anywhere. My longer-term purpose is to continue pushing Naptime towards not being completely tied to HTTP, and changing access control to not exclusively be about `RequestHeader` is part of that. I'm imagining that the `Rest.auth(...)` DSL will be expanded to handle general `AccessControl` rather than just `HeaderAccessControl` (which I think will have more consequences than just this PR on how access controls are implemented).

naptime/src/main/scala/org/coursera/naptime/access/AccessControl.scala
josh-newman: Hm, interesting. I was thinking that the authentication step is what's actually being "simulated" because it's faked here, but I guess it's reasonable to interpret "simulation" as "run in a fake environment". What do you think of calling this `runWithFakeAuthentication`?

naptime/src/main/scala/org/coursera/naptime/access/AccessControl.scala
saeta: I definitely like the goal of abstracting away from HTTP. I guess I'm not convinced leaving an unbounded, contravariant type parameter is the right way to go. I'd rather defer introducing this trait until after we have a better sense on what interface we're actually trying to provide in lieu of Play-HTTP. (i.e. no current uses of `HeaderAccessControl` can migrate to `AccessControl`.)

Additionally, I'm not sure I like the name of the trait either. :-( There's a lot more to access control than Auth-n/z at the request level. There's field-level access control, and response element-level access control. I don't think these sorts of access control applications would fit under the same interface.

naptime/src/main/scala/org/coursera/naptime/access/AccessControl.scala
saeta: I like `runWithFakeAuthentication` the best. +1

PR #64, opened 2016-07-19
Proposal automatic resource inclusion abstractions

saeta:
Define key abstractions that will form the basis for the automatic
resource inclusion and GraphQL work.

### DIFF
--- a/naptime/src/main/scala/org/coursera/naptime/ari/models.scala
+++ b/naptime/src/main/scala/org/coursera/naptime/ari/models.scala
@@ -1,0 +1,95 @@
+package org.coursera.naptime.ari
+
+import com.linkedin.data.DataMap
+import org.coursera.naptime.ResourceName
+import org.coursera.naptime.ResponsePagination
+import org.coursera.naptime.schema.Resource
+import play.api.libs.json.JsValue
+import play.api.mvc.RequestHeader
+
+import scala.concurrent.Future
+import scala.collection.immutable
+
+/**
+ * The engine layer presents this EngineAPI to the presentation layer. The engine layer handles query validation,
+ * execution planning & optimization, and finally orchestrates the fetching of data by leveraging the [[FetcherApi]].
+ *
+ * Engine's must be threadsafe as they are called from multiple threads simultaneously.
+ */
+trait EngineApi {
+  /**
+   * Fetches all data needed to formulate a response to the query.
+   *
+   * TODO: Consider returning an `Either[Set[Errors], Future[Response]]` instead for more structured validation errors
+   *
+   * @param request The resource request to execute.
+   * @return A future of all available data needed to construct a response to the query. It is up to the presentation
+   *         layer to turn that into the response format as required.
+   */
+  def execute(request: Request): Future[Response]
+
+  /**
+   * Gets the Naptime schema currently in use by the presentation layer.
+   *
+   * TODO: include the Courier model schemas
+   */
+  def schema: Seq[Resource]
+}
+
+/**
+ * The fetcher calls the Naptime APIs (either local or remote) to acquire all of the data necessary.
+ */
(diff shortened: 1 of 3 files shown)


naptime/src/main/scala/org/coursera/naptime/ari/models.scala
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * @param requestHeader The request header is used for authentication parsing in the underlying requests made. Path
+ *                      and query parameters included in the request header are ignored in favor of the data within the
+ *                      topLevelRequests fields.
+ * @param topLevelRequests A non-empty list of "roots" used to begin queries. Every query begins with a "base" resource
+ *                         and related models are joined upon it from there.
+ */
+case class Request(
+  requestHeader: RequestHeader,
+  topLevelRequests: Seq[(ResourceName, RequestField)]
```
bryan-coursera: should we turn this into a case class so we don't have to rely on remembering the order of these parameters here? also, what are your thoughts on making `RequestField` a trait that can be extended by `TopLevelRequestField` which has an additional ResourceName parameter? that way, in any helper methods or other places, we can treat this request just like a regular field without having to do any extra parsing

naptime/src/main/scala/org/coursera/naptime/ari/models.scala
```diff
@@ -0,0 +1,87 @@
...
+ * @param name The name of the requested field.
+ * @param alias The name the field should be renamed to in the response.
+ * @param args If the field takes parameters, they are encapsulated here.
+ * @param relatedResource The list of fields in the related resource.
+ */
+case class RequestField(
+  name: String,
+  alias: Option[String],
+  args: Set[(String, JsValue)],
+  relatedResource: List[RequestField])
```
bryan-coursera: in graphql, these are called "selections", which i think better expresses that these fields are sub-selections on this field. thoughts?

naptime/src/main/scala/org/coursera/naptime/ari/models.scala
```diff
@@ -0,0 +1,87 @@
...
+ * @param output A map from resource name to a response
+ */
+case class Response(
+  output: Map[ResourceName, ResourceResponse]
+)
+
+// TODO: figure out directives
+
+// TODO: key may change to a more sophisticated type. It should be a ResourceName, but with hydrated path parameters.
+case class ResourceResponse(key: String, models: Seq[DataMap])
```
bryan-coursera: should this also include pagination / other response metadata?

naptime/src/main/scala/org/coursera/naptime/ari/models.scala
saeta: Because you can't have a case class extend from a case class, I felt it'd be better to do it this way. I'll turn the tuple `(ResourceName, RequestField)` into a case class, however.

saeta: In refactoring to a case class, I realized that this should instead be: `Seq[(ResourceName, Seq[RequestField])]`, so it definitely doesn't make sense making `RequestField` into a trait, because it's a slightly different structure.

naptime/src/main/scala/org/coursera/naptime/ari/models.scala
saeta: I like that better. Changed.

naptime/src/main/scala/org/coursera/naptime/ari/models.scala
saeta: definitely include pagination. the question is: should it include an `ETag` in some fashion.

bryan-coursera:
i'm not thrilled about the package name, but let's keep brainstorming and consider changing it in the future. ignoring that though, LGTM

PR #65, opened 2016-07-20
Relax type restrictions for actions.

saeta:
Previously Engine2's required that the response object in actions be
serializable to a `DataMap`. This is too restrictive, because actions
are the escape hatches from the Naptime system, and must thus support
responses of strings, `JsValues`, and other primitive types. We thus
define a new type that is more general and more flexible to support
actions.


zhaojunz:
Lgtm

PR #69, opened 2016-07-22
Add initial GraphQL parsing library

bryan-coursera:
- Adds sangria dependency to Naptime
- Adds a GraphQL Parser object that can parse a GraphQL String into a Naptime Request object
- Recursively parses fields into Naptime RequestFields
- Adds unit tests for parsing

Still a few TODOs left around error handling and argument parsing (GraphQL AST types into JSON)

PTAL @saeta


saeta:
LGTM.

PR #70, opened 2016-07-22
Generalize Naptime Action Serializer's Play-JSON

saeta:
We can support serializing any object that defines a Play-JSON Writes.


zhaojunz:
lgtm thanks!

PR #72, opened 2016-07-27
Incorporate related resource information in schema

saeta:
Summary:
In order to build the automatic resource inclusion engine, to support
GraphQL, and add links to the documentation between related resources,
we need to incorporate the related fields information into the "merged"
type schema.

Because the Engine, GraphQL and others are centered around Courier and
Pegasus schemas, I've begun constructing a set of classes that can form
the basis for a more integration-style test suite in the
`NestedMacroCourierTests` class. The first test in this class is a test
of the related fields information functionality added in this commit.

Because we are centering upon courier models, I've added a couple extra
abstract classes that reduce the amount of boilerplate required for
resources using courier types.

### DIFF
--- a/naptime-testing/src/test/scala/org/coursera/naptime/NestedMacroCourierTests.scala
+++ b/naptime-testing/src/test/scala/org/coursera/naptime/NestedMacroCourierTests.scala
@@ -1,0 +1,107 @@
+package org.coursera.naptime
+
+import com.linkedin.data.schema.DataSchema
+import com.linkedin.data.schema.RecordDataSchema
+import org.coursera.naptime.couriertests.ExpectedMergedCourse
+import org.coursera.naptime.model.Keyed
+import org.coursera.naptime.resources.CourierCollectionResource
+import org.coursera.naptime.router2.Router
+import org.junit.Test
+import org.scalatest.junit.AssertionsForJUnit
+
+/**
+ * This test suite uses Courier to exercise advanced use cases for Naptime.
+ */
+object NestedMacroCourierTests {
+
+  class CoursesResource extends CourierCollectionResource[String, Course] {
+    override def resourceName: String = "courses"
+
+    override implicit lazy val Fields: Fields[Course] = BaseFields.withRelated("instructors" -> ResourceName("instructors", 1))
+
+    /**
+     * This `var` can be overridden to help fake out the implementation in particular functions.
+     */
+    var relatedFunction: Option[String => List[String]] = None
+
+    private[this] def fallbackRelatedFunction(id: String): List[String] = {
+      id match {
+        case "abc" => List("123")
+        case "xyz" => List.empty
+        case "qrs" => List("456", "789")
+        case _ => List.empty
+      }
+    }
+
+    private[this] def makeCourse(id: String): Keyed[String, Course] = {
+      val relatedInstructors = relatedFunction.getOrElse(fallbackRelatedFunction _)(id)
+
+      Keyed(id, Course(s"course-$id", s"$id description", relatedInstructors))
+    }
+
+    def get(id: String) = Nap.get { ctx =>
+      Ok(makeCourse(id))
+    }
+
+    def multiGet(ids: Set[String]) = Nap.multiGet { ctx =>
+      Ok(ids.map(id => makeCourse(id)).toList)
+    }
+
+    def search(query: Option[String]) = Nap.finder { ctx =>
+      val ids = query.map(List(_)).getOrElse(List("abc", "qrs"))
+      Ok(ids.map(id => makeCourse(id)))
+    }
+
+    def byInstructor(instructorId: String) = Nap.finder { ctx =>
+      val ids = instructorId match {
+        case "123" => List("xyz")
+        case "456" => List("xyz", "abc")
+        case "789" => List("qrs")
+        case _ => List.empty
+      }
+      Ok(ids.map(id => makeCourse(id)))
+    }
+  }
+
+  class InstructorsResource extends CourierCollectionResource[String, Instructor] {
+    override def resourceName: String = "instructors"
+
+    def multiGet(ids: Set[String]) = Nap.multiGet { ctx =>
+      ???
+    }
+
+  }
+
+  val courseRouter = Router.build[CoursesResource]
+  val instructorRouter = Router.build[InstructorsResource]
+}
+
+class NestedMacroCourierTests extends AssertionsForJUnit {
+
+  @Test
+  def checkCoursesMergedType(): Unit = {
+    val types = NestedMacroCourierTests.courseRouter.types
+    assert(3 === types.size, s"Got $types")
+    val mergedTypeOpt = types.find(_.key == "org.coursera.naptime.NestedMacroCourierTests.CoursesResource.Model")
+    assert(mergedTypeOpt.isDefined)
+    val mergedType = mergedTypeOpt.get
+    assert(!mergedType.value.hasError)
+    assert(mergedType.value.getType === DataSchema.Type.RECORD)
+    val mergedValueRecord = mergedType.value.asInstanceOf[RecordDataSchema]
+    assert(4 === mergedValueRecord.getFields.size(), mergedValueRecord)
+    val instructorsField = mergedValueRecord.getField("instructors")
+    assert(null != instructorsField, instructorsField)
+    assert(null != instructorsField.getProperties)
+    val typesProperty = instructorsField.getProperties.get(Types.Relations.PROPERTY_NAME)
+    assert(null != typesProperty)
+    assert(typesProperty.isInstanceOf[String])
+    assert(typesProperty === "instructors.v1")
+  }
+
+  @Test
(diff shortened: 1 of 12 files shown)


naptime-testing/src/test/scala/org/coursera/naptime/NestedMacroCourierTests.scala
```diff
@@ -0,0 +1,95 @@
...
+ * This test suite uses Courier to exercise advanced use cases for Naptime.
+ */
+object NestedMacroCourierTests {
+
+  class CoursesResource extends TopLevelCourierCollectionResource[String, Course] {
+    override def resourceName: String = "courses"
+
+    override implicit lazy val Fields: Fields[Course] = BaseFields.withRelated("instructors" -> ResourceName("instructors", 1))
+
+    var relatedFun: Option[String => List[String]] = None
```
bryan-coursera: small nit, you use `fallbackRelatedFunction` below but abbreviate `fun` here. i'd recommend using `function` here too to be consistent.

I also don't follow the need for having a relatedFunction helper and a fallback -- it doesn't look like these helpers are overriding anything in a base class, so i'm not positive why there's even a fallback.

naptime-testing/src/test/scala/org/coursera/naptime/NestedMacroCourierTests.scala
```diff
@@ -0,0 +1,95 @@
...
+        case "abc" => List("123")
+        case "xyz" => List.empty
+        case "qrs" => List("456", "789")
+        case _ => List.empty
+      }
+    }
+
+    private[this] def makeCourse(id: String): Keyed[String, Course] = {
+      Keyed(id,
+        Course(s"course-$id", s"$id description", relatedFun.map(_(id)).getOrElse(fallbackRelatedFunction(id))))
```
bryan-coursera: style nit: i'd prefer
```scala
relatedFun.getOrElse(fallbackRelatedFunction)(id)
```
to only have to invoke it once

naptime-testing/src/test/scala/org/coursera/naptime/NestedMacroCourierTests.scala
saeta: I made this a `var` instead of a `val` so that tests could optionally configure different behavior. +1 about making the names consistent, however.

Naptime's style doesn't work so well with Mockito / other mocking libraries. So this is sort of a fake implementation.

bryan-coursera: ah ok. can you just add a brief comment explaining that?

naptime/src/main/scala/org/coursera/naptime/Types.scala
```diff
@@ -58,6 +64,19 @@ object Types extends StrictLogging {
...
     }
+    for ((name, relation) <- fields.relations) {
+      Option(mergedSchema.getField(name)) match {
+        case None =>
+          logger.warn(s"Fields for resource $typeName mentions field name '$name' but this field " +
+            "is not found in the merged type schema.")
+        case Some(field) =>
+          val properties = field.getProperties.asScala
+          val relatedMap = Map[String, AnyRef](
+            Relations.PROPERTY_NAME -> relation.identifier)
```
bryan-coursera: do we validate these identifiers anywhere (ensuring they're at least a valid resource name)?

naptime-testing/src/test/scala/org/coursera/naptime/NestedMacroCourierTests.scala
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  val courseRouter = Router.build[CoursesResource]
+  val instructorRouter = Router.build[InstructorsResource]
+}
+
+class NestedMacroCourierTests extends AssertionsForJUnit {
+
+  @Test
+  def checkCoursesMergedType(): Unit = {
```
bryan-coursera: this test looks good. I wonder if there's any way though to write an expected courier model, and assert that the generated one and the hard-coded one are equal?

that way you could just write something like:
```
record MergedCourse {

  id: String

  @related("instructors.v1")
  instructors: [InstructorId]

  ...

}
```
and somehow assert that that is equal to the generated one.  it might make the tests more maintainable and easier to comprehend what's going on

naptime/src/main/scala/org/coursera/naptime/resources/resources.scala
```diff
@@ -165,3 +169,58 @@ trait CollectionResource[ParentResource <: Resource[_], K, M] extends Resource[M
...
+ * some boilerplate.
+ *
+ * If you have a nested resource, check out [[CourierCollectionResource]].
+ *
+ * @param kf The [[KeyFormat]] for the key type in the resource.
+ * @param classTag The concrete class tag for the value type in the resource.
+ * @tparam K The key type of the resource.
+ * @tparam M The "value" type of the resource.
+ */
+abstract class TopLevelCourierCollectionResource[K, M <: ScalaRecordTemplate](
```
bryan-coursera: can we make `TopLevelCourierCollectionResource` the default `CourierCollectionResource`, and either have a `NestedCourierCollectionResource` or some other way to override the `parentResource`? It seems like the top-level one is going to be used 99% of the time.

bryan-coursera:
this looks pretty good to me, but i'd recommend having @josh-newman check out the `CourierCollectionResource` helpers, because i'm less familiar with the resource code in general.

naptime-testing/src/test/scala/org/coursera/naptime/NestedMacroCourierTests.scala
saeta: done.

naptime-testing/src/test/scala/org/coursera/naptime/NestedMacroCourierTests.scala
saeta: done.

naptime/src/main/scala/org/coursera/naptime/Types.scala
saeta: We should after we aggregate them all together in production. Until then, we don't have the complete picture. :-(

naptime/src/main/scala/org/coursera/naptime/resources/resources.scala
saeta: Excellent point. Done.

naptime-testing/src/test/scala/org/coursera/naptime/NestedMacroCourierTests.scala
saeta: I just pushed a new test that does roughly this. Thanks for suggesting it.

naptime-testing/src/test/scala/org/coursera/naptime/NestedMacroCourierTests.scala
```diff
@@ -0,0 +1,107 @@
...
+      id match {
+        case "abc" => List("123")
+        case "xyz" => List.empty
+        case "qrs" => List("456", "789")
+        case _ => List.empty
+      }
+    }
+
+    private[this] def makeCourse(id: String): Keyed[String, Course] = {
+      val relatedInstructors = relatedFunction.getOrElse(fallbackRelatedFunction _)(id)
```
bryan-coursera: oh, i forgot about the need to add that underscore in there to partially apply it. this looks uglier than i had hoped :/. sorry!

naptime-testing/src/test/scala/org/coursera/naptime/NestedMacroCourierTests.scala
```diff
@@ -0,0 +1,107 @@
...
+    assert(null != typesProperty)
+    assert(typesProperty.isInstanceOf[String])
+    assert(typesProperty === "instructors.v1")
+  }
+
+  @Test
+  def checkMergedCourseModelSchema(): Unit = {
+    val types = NestedMacroCourierTests.courseRouter.types
+    val mergedType = types.find(_.key == "org.coursera.naptime.NestedMacroCourierTests.CoursesResource.Model").get
+    assert(ExpectedMergedCourse.SCHEMA.getFields === mergedType.value.asInstanceOf[RecordDataSchema].getFields)
```
bryan-coursera: this is awesome. thanks!

naptime/src/main/scala/org/coursera/naptime/Types.scala
bryan-coursera: hmm, alright. once its on the courier model, would we take it back off? (if we're going to make decisions just based on the schema, i hope that's not too late)

naptime-testing/src/test/scala/org/coursera/naptime/NestedMacroCourierTests.scala
saeta: I initially wrote the test to compare everything, but the namespaces are off, as well as the type names, so I just check field equality, but I still think that's a step in the right direction. :-)

PR #75, opened 2016-08-01
Add schema generation for GraphQL

bryan-coursera:
Creates a helper class that converts Pegasus schemas and Naptime resource definitions into a full GraphQL schema.


bryan-coursera:
PTAL @saeta

saeta:
LGTM.

PR #76, opened 2016-08-01
ARI: Initial Engine Impl Tests

saeta:
Set up the basic infrastructure for testing the engine implementation.
Currently, we only have a trivial engine implementation that cannot
handle any automatic joining.

### DIFF
--- a/naptime/src/main/scala/org/coursera/naptime/ari/engine/EngineImpl.scala
+++ b/naptime/src/main/scala/org/coursera/naptime/ari/engine/EngineImpl.scala
@@ -1,0 +1,20 @@
+package org.coursera.naptime.ari.engine
+
+import com.linkedin.data.schema.DataSchema
+import org.coursera.naptime.ari.EngineApi
+import org.coursera.naptime.ari.FetcherApi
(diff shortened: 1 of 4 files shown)


saeta:
PTAL @bryan-coursera

naptime/src/main/scala/org/coursera/naptime/ari/engine/EngineImpl.scala
```diff
@@ -0,0 +1,20 @@
...
+import org.coursera.naptime.ari.EngineApi
+import org.coursera.naptime.ari.FetcherApi
+import org.coursera.naptime.ari.Request
+import org.coursera.naptime.ari.Response
+import org.coursera.naptime.schema.Resource
+
+import scala.concurrent.Future
+
+class EngineImpl(
+    override val schemas: Seq[Resource],
```
bryan-coursera: i've been using naming where this would be called `resources` and the map of string->dataschema would be called `schemas`. i'm fine with either, but let's just make sure to be consistent.

bryan-coursera:
let's clean up some of the fake data, but otherwise LGTM

PR #79, opened 2016-08-11
Design response model + EngineImpl improvements

saeta:
This commit implements the long term response model, as per the design
discussions. It converts all ARI-related code to use it, and adds a
number of tests to ensure correctness.

Additionally, thanks to the new response model, we are able to implement
a more sophisticated handling of multiple top-level requests in the
engine.

### DIFF
--- a/naptime/src/main/scala/org/coursera/naptime/ari/models.scala
+++ b/naptime/src/main/scala/org/coursera/naptime/ari/models.scala
@@ -1,15 +1,14 @@
 package org.coursera.naptime.ari
 
+import com.linkedin.data.DataList
 import com.linkedin.data.DataMap
 import com.linkedin.data.schema.DataSchema
 import org.coursera.naptime.ResourceName
-import org.coursera.naptime.ResponsePagination
 import org.coursera.naptime.schema.Resource
 import play.api.libs.json.JsValue
 import play.api.mvc.RequestHeader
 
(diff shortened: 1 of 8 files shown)


naptime/src/main/scala/org/coursera/naptime/ari/models.scala
```diff
@@ -79,21 +80,49 @@ case class TopLevelRequest(resource: ResourceName, selection: RequestField)
...
+    val mergedTopLevel = topLevelIds ++ other.topLevelIds
+    val mergedData = (data.keySet ++ other.data.keySet).map { resourceName =>
+      val lhs = data.getOrElse(resourceName, Map.empty)
+      val rhs = other.data.getOrElse(resourceName, Map.empty)
+      val mergedMap = (lhs.keySet ++ rhs.keySet).map { key =>
+        val lh = lhs.get(key)
+        val rh = rhs.get(key)
+        val merged = (lh, rh) match {
+          case (None, None) =>
+            throw new IllegalStateException("This should never occur?!?!")
```
bryan-coursera: better message here please

bryan-coursera:
LGTM

naptime/src/main/scala/org/coursera/naptime/ari/models.scala
saeta: done.

PR #80, opened 2016-08-12
Support basic joins.

saeta:
This commit adds support for a single level of automated nested/joining
of related resources. Subsequent commits will add support for more
foreign-key types of joins, as well as support alternate joins such
as reverse includes / etc.

Additionally, this commit cleans up the engine API. There were unused
components of the engine API that won't be used. By removing the
unnecessary additional fields in the engine API, we avoid tangling the
code and subsequent developer confusion.

### DIFF
--- a/naptime/src/main/scala/org/coursera/naptime/ari/engine/EngineImpl.scala
+++ b/naptime/src/main/scala/org/coursera/naptime/ari/engine/EngineImpl.scala
@@ -2,42 +2,108 @@ package org.coursera.naptime.ari.engine
 
 import javax.inject.Inject
 
-import com.linkedin.data.schema.DataSchema
+import com.linkedin.data.schema.RecordDataSchema
+import com.typesafe.scalalogging.StrictLogging
+import org.coursera.naptime.ResourceName
+import org.coursera.naptime.Types.Relations
 import org.coursera.naptime.ari.EngineApi
 import org.coursera.naptime.ari.FetcherApi
 import org.coursera.naptime.ari.Request
+import org.coursera.naptime.ari.RequestField
 import org.coursera.naptime.ari.Response
+import org.coursera.naptime.ari.TopLevelRequest
 import org.coursera.naptime.router2.NaptimeRoutes
 import org.coursera.naptime.schema.Resource
+import play.api.libs.json.JsString
 
+import scala.collection.JavaConverters._
 import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 
 class EngineImpl @Inject() (
     naptimeRoutes: NaptimeRoutes,
     fetcher: FetcherApi)
-    (implicit executionContext: ExecutionContext) extends EngineApi {
+    (implicit executionContext: ExecutionContext) extends EngineApi with StrictLogging {
 
-  override val schemas: Seq[Resource] = naptimeRoutes.routerBuilders.map(_.schema).toList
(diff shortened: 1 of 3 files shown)


naptime/src/main/scala/org/coursera/naptime/ari/engine/EngineImpl.scala
```diff
@@ -2,42 +2,108 @@ package org.coursera.naptime.ari.engine
...
+    topLevelResponse.flatMap { topLevelResponse =>
+      // If we have a schema, we can perform automatic resource inclusion.
+      (for {
+        topLevelSchema <- resourceSchemas.get(topLevelRequest.resource)
+        mergedType <- mergedTypes.get(topLevelSchema.mergedType)
+      } yield {
+        val topLevelData = topLevelResponse.data.get(topLevelRequest.resource).toIterable.flatMap { dataMap =>
+          dataMap.values
+        }
+        val nestedLookups = topLevelRequest.selection.selections.filter(_.selections.nonEmpty)
```
bryan-coursera: I don't know if we can immediately assume that a subselection means an related resource.
How do we plan to handle the case where we have a schema like the following:
```
record CourseDetails {
  startDate: String
  endDate: String
}
record Course {
  details: CourseDetails
}
```

would you still reference the fields like:
```
query {
  CoursesV1Resource {
    get(id: "v1-123") {
      courseDetails {
        startDate
      }
    }
  }
}
```

naptime/src/main/scala/org/coursera/naptime/ari/engine/EngineImpl.scala
```diff
@@ -2,42 +2,108 @@ package org.coursera.naptime.ari.engine
...
+                throw new IllegalStateException(s"Unexpected type for identifier '$identifier' for field '$field' " +
+                  s"in merged type $mergedType")
+            }
+          }.getOrElse {
+            throw new IllegalStateException(s"Could not find nested field $nestedField on merged type $mergedType")
+          }
+
+          val multiGetIds = topLevelData.flatMap { elem =>
+            // TODO: support single foreign key relations (currently this supports only nested lists)
+            val field = Option(elem.getDataList(nestedField.name)).map(_.asScala).getOrElse {
```
bryan-coursera: what if we have a one-to-one relationship (a course has a single courseDetails.v1 related thing) -- it looks like you're only supporting datalists here?

bryan-coursera:
a couple of questions, but other than those concerns the code looks good

naptime/src/main/scala/org/coursera/naptime/ari/engine/EngineImpl.scala
saeta: Ah, you're quite right. I previously forgot about that. I think instead of throwing on line 74, I should handle it appropriately. I'll add a test and then the fix in a subsequent PR.

naptime/src/main/scala/org/coursera/naptime/ari/engine/EngineImpl.scala
saeta: Yeah, thats what the TODO on the previous line (82) is about.

saeta:
Thanks for the super speedy review. I'm going to start working on subsequent PRs to build upon this and fix some of the TODOs in there.

naptime/src/main/scala/org/coursera/naptime/ari/engine/EngineImpl.scala
bryan-coursera: haha, i see that now :)

PR #82, opened 2016-08-16
Add multi-join test.

saeta:
Add a test to verify the engine correctly handles joining multiple
resources on to a top level resource.

Additionally, this commit reorganizes all the TODOs into a single list
in the EngineImplTest class.


bryan-coursera:
lgtm

PR #83, opened 2016-08-17
Use fully-qualified reference to Keyed

bryan-coursera:
The latest version of naptime causes the following error when compiling:
```
[error] /Users/bryan/base/coursera/infra-services/services/rapidash/app/org/coursera/rapidash/RapidashModule.scala:40: not found: value Keyed
[error]     bindResource[AppResource]
```

Using the full reference to `Keyed` should resolve this.

PTAL @saeta . cc @yifan-coursera


saeta:
Thank you @bryan-coursera This is perfect.

PR #84, opened 2016-08-19
Fix computing full path.

saeta:
Computing the full path is required to display the correct documentation.
In order to handle nested resources correctly, we need to stop recursing
when we hit the root resource. This commit fixes that.


bryan-coursera:
LGTM

PR #87, opened 2016-08-22
Update macro to support primitive defaults and misc. fixes

bryan-coursera:
Updates the naptime resource macro to support defaults of a primitive type (instead of just record types), and wraps some potentially dangerous operations in schema building in a try / catch until @saeta gets schema models in a better state


saeta:
LGTM.

PR #88, opened 2016-08-23
Clean up some code

bryan-coursera:
PTAL @saeta. This just cleans a few things up and adds some documentation, so I'm less ashamed of the code.


saeta:
LGTM. Thanks for doing this!

PR #89, opened 2016-08-26
Handle more edge cases in schema generation

bryan-coursera:
This handles a few more edge cases around things being null or having previously unsupported types.

PTAL @saeta


saeta:
LGTM.

PR #91, opened 2016-08-31
Separate graphql/sangria dependencies to own lib

saeta:
In order to avoid polluting the classpath for naptime users that are not
interested in sangria and/or graphql, we have split up those
dependencies into a separate library called `naptime-graphql`.


version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.2.3"
+version in ThisBuild := "0.2.3-graphql"
```
saeta: In order to get sbt to resolve the new build configuration appropriately, it's critical that the version is not in the local ivy cache. I've appended this suffix in order to get it to resolve properly. In the future, I'll look into something like `sbt-release` to automate the version bumping / etc. (and including a `-dev` suffix or some-such).

bryan-coursera:
LGTM. I'm assuming naptime itself can compile without the graphql/sangria stuff loaded at all, right?

Thanks for doing this!

saeta:
Yup, it can compile + test without any sangria in it.

PR #95, opened 2016-09-02
Convert naptime defaults to use ArbitraryValue courier type

bryan-coursera:
This augments the "default" value for parameters in a REST action to include additional information about types. Currently, the default value is just the string representation of the value. However, because external clients don't have information about the resource schema, its impossible to know what type the default value is.

By using a Courier union type, we can include "metadata" about the default value's type, which can then be parsed by other servers more easily.


saeta:
LGTM. Thanks for bringing this from `graphql` into `master`.

PR #97, opened 2016-09-10
Handle recursive model definitions and null field definitions

bryan-coursera:
This handles two different cases, and fixes schema generation:
- Field definitions that are `NullType`
- Field definitions that contain recursive references

PTAL @saeta


saeta:
LGTM.

PR #98, opened 2016-09-12
Infer additional key type schemas

saeta:
It is quite common to use a more sophisticated key type. Often, these key types
have a single apply method in the companion object that corresopnds to the
fields of the serialized form of the type. This commit extends the macro to
automatically infer the schema for these more sophisticated key types.

Additionally, we capture the changes in some unit tests to verify the correct
operation of the macro.


bryan-coursera:
I didn't fully understand some of the macro stuff, but your tests look sufficient, so LGTM

PR #99, opened 2016-09-12
Implement handling for typerefs in Types.scala

saeta:
Note: I'm not sure that this code path will ever be hit in reality,
due to the constraints placed on schemas in other parts of the
framework.


bryan-coursera:
LGTM

PR #100, opened 2016-09-12
Wrap the NPE thrown by DataMap to include info.

saeta:
Currently, when a null field is encountered by the serializer, the exception
thrown gives no details as to what field was actually null.

We convert it to a RuntimeException, because java.lang.NPE doesn't include
a 2-arg constructor allowing an underlying exception to be passed through.


bryan-coursera:
LGTM

PR #102, opened 2016-09-13
Handle Key types that define their own id field

saeta:
In some cases, a key courier type might define its own id field (e.g.
only a single part of the key is required to uniquely identify the
record). This commit makes the Types computation a little bit smarter.

Note that relying on this behavior is a potential code smell.

### DIFF
--- a/naptime/src/main/scala/org/coursera/naptime/Types.scala
+++ b/naptime/src/main/scala/org/coursera/naptime/Types.scala
@@ -114,12 +114,16 @@ object Types extends StrictLogging {
       valueType: RecordDataSchema): RecordDataSchema = {
     val errorMessageBuilder = new StringBuilder
     val recordDataSchema = new RecordDataSchema(new Name(typeName), RecordType.RECORD)
-    val fieldsMinusId = keyType.getFields().asScala.toList ++ valueType.getFields().asScala
-    val idField = new RecordDataSchema.Field(new StringDataSchema)
-    idField.setOptional(false)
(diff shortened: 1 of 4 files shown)


naptime/src/main/scala/org/coursera/naptime/Types.scala
```diff
@@ -115,11 +115,15 @@ object Types extends StrictLogging {
     val errorMessageBuilder = new StringBuilder
     val recordDataSchema = new RecordDataSchema(new Name(typeName), RecordType.RECORD)
     val fieldsMinusId = keyType.getFields().asScala.toList ++ valueType.getFields().asScala
-    val idField = new RecordDataSchema.Field(new StringDataSchema)
-    idField.setOptional(false)
-    idField.setName("id", errorMessageBuilder)
-    idField.setRecord(recordDataSchema)
-    val fields = idField :: fieldsMinusId
+    val fields = if (fieldsMinusId.exists(_.getName == "id")) {
+      fieldsMinusId // The key type had an `id` field defined as part of it.
```
bryan-coursera: i'm not crazy about the name `fieldsMinusId`, since its not really fields minus id. Can we just call this "combinedFields" instead maybe?

bryan-coursera:
LGTM, other than the one naming nit. I also agree that we shouldn't let fields define their own id, or maybe _only_ permit it if the id field is the only field in the key record type? But let's chat about this at some other point -- maybe make a ticket for this now to make sure we don't forget it in the future?

PR #103, opened 2016-09-14
Set default fields for empty object types, and force all id fields to be strings

bryan-coursera:
GraphQL doesn't allow ObjectTypes to not have any fields, which was causing errors in GraphiQL (but not Sangria?). This adds a "fake" field to empty objects for now, until we can further diagnose why these are empty. The long term strategy would probably be to not include these types if they're empty though -- we should chat more.

This also forces all "id" fields to be of string type, which seems to be a naptime thing for merged models.

PTAL @saeta


saeta:
@bryan-coursera why force all ID fields to be strings?

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/SangriaGraphQlSchemaBuilder.scala
```diff
@@ -323,27 +330,34 @@ class SangriaGraphQlSchemaBuilder(
...
-      case arrayField: ArrayDataSchema => context => context.value.getDataList(fieldName)
-      case recordField: RecordDataSchema => context => context.value.getDataMap(fieldName)
+        getSangriaResolverForSchema(typerefField.getDereferencedDataSchema, fieldName)
       case _ =>
-        logger.error(s"Could not match schema type $schemaType")
-        context => null
+        val baseResolver: (Context[SangriaGraphQlContext, DataMap]) => Any = {
+          schemaType match {
+            case stringField: StringDataSchema => context => context.value.getString(fieldName)
+            case _ if fieldName == "id" => context => context.value.getString(fieldName)
```
saeta: Why do you do this?

bryan-coursera:
Test case changes look reasonable to me, so if the rest of the changes are good for you, feel free to merge it in

PR #104, opened 2016-09-15
Abstract NaptimeRoutes behind SchemaProvider trait.

saeta:
Previously, the EngineImpl directly pulled from the local naptime resources.
However, this does not work in distributed mode. Instead, we must add a layer
of indirection to allow for custom distributed / remote SchemaProviders.


bryan-coursera:
LGTM. Thanks!

PR #106, opened 2016-09-19
Fix parsing of complex child types in lists

bryan-coursera:
List types weren't being parsed correctly (due to the missing `.asScala`), causing a parsing exception and the entire query for any field that was a list to fail. This fixes that issue, and adds a simple test.

PTAL @saeta


bryan-coursera:
I'll merge it in when we're ready to cut 0.3.1

PR #107, opened 2016-09-20
Include GraphQL Play controller in OSS Naptime.

saeta:
Additionally, this commit refactors a few things to make supporting
a constantly changing underlying naptime schema easier to support.

### DIFF
--- a/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/GraphQLController.scala
+++ b/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/GraphQLController.scala
@@ -1,0 +1,91 @@
+package org.coursera.naptime.ari.graphql
+
+import javax.inject._
+
+import com.typesafe.scalalogging.StrictLogging
+import org.coursera.naptime.ari.EngineApi
+import org.coursera.naptime.ari.Response
+import play.api.libs.json.JsObject
+import play.api.libs.json.JsString
+import play.api.libs.json.Json
+import play.api.mvc._
+import sangria.execution.ErrorWithResolver
+import sangria.execution.Executor
+import sangria.execution.QueryAnalysisError
+import sangria.parser.QueryParser
+import sangria.renderer.SchemaRenderer
+import sangria.marshalling.playJson._
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+
+
+@Singleton
+class GraphQLController @Inject() (
+    graphqlSchemaProvider: GraphqlSchemaProvider,
+    schemaProvider: GraphqlSchemaProvider,
+    engine: EngineApi)
+    (implicit ec: ExecutionContext)
+  extends Controller
+    with StrictLogging {
+
+
+  def renderSchema = Action {
+    Ok(SchemaRenderer.renderSchema(graphqlSchemaProvider.schema))
+  }
+
+  def graphqlBody = Action.async(parse.json) { request =>
+    val query = (request.body \ "query").as[String]
+    val operation = (request.body \ "operationName").asOpt[String]
+
+    val variables = (request.body \ "variables").toOption.flatMap {
+      case JsString(vars) => Some(parseVariables(vars))
+      case obj: JsObject => Some(obj)
+      case _ => None
+    }
+
+    executeQuery(query, request, variables, operation)
+  }
+
+  private def parseVariables(variables: String) = {
+    if (variables.trim == "" || variables.trim == "null") {
+      Json.obj()
+    } else {
+      Json.parse(variables).as[JsObject]
+    }
+  }
+
+  private def executeQuery(
+      query: String,
+      requestHeader: RequestHeader,
+      variables: Option[JsObject],
+      operation: Option[String]) = {
+
+    // TODO(bryan): Handle errors / failures properly here
+    (for {
+      request <- SangriaGraphQlParser.parse(query, requestHeader)
+      document <- QueryParser.parse(query).toOption
+    } yield {
+      val fetcherExecution: Future[Option[Response]] = if (query.contains("IntrospectionQuery")) {
+        Future.successful(None)
+      } else {
+        engine.execute(request).map(Some(_))
+      }
+      fetcherExecution.flatMap { responseOpt =>
+        val allData = responseOpt.map(
+          _.data.map { case (resourceName, response) =>
+            resourceName.identifier -> response.values.toList
+          }).getOrElse(Map.empty)
+        val context = SangriaGraphQlContext(data = allData)
+        Executor.execute(graphqlSchemaProvider.schema, document, context)
+          .map(Ok(_))
+          .recover {
+            case error: QueryAnalysisError => BadRequest(error.resolveError)
+            case error: ErrorWithResolver => InternalServerError(error.resolveError)
+          }
+      }
+    }).getOrElse {
+      Future.successful(BadRequest("Invalid request"))
+    }
+  }
+}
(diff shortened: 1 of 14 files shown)


naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/GraphQLController.scala
```diff
@@ -0,0 +1,92 @@
+package org.coursera.naptime.ari.graphql
```
bryan-coursera: Small nit: do we need to include a license at the top of all of the files? (I think I've been missing that too)

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/GraphQLController.scala
```diff
@@ -0,0 +1,92 @@
...
+import sangria.renderer.SchemaRenderer
+import sangria.marshalling.playJson._
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+
+
+@Singleton
+class GraphQLController @Inject() (
+  graphqlSchemaProvider: GraphqlSchemaProvider,
```
bryan-coursera: if we're following our coding style, then this should be 4 space indented, and `withStrictLogging` should be two-space indented

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/GraphQLController.scala
```diff
@@ -0,0 +1,92 @@
...
+  schemaProvider: GraphqlSchemaProvider,
+  engine: EngineApi)
+  (implicit ec: ExecutionContext)
+  extends Controller
+    with StrictLogging {
+
+
+  def index = Action {
```
bryan-coursera: let's rename this to renderSchema

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/GraphQLController.scala
```diff
@@ -0,0 +1,92 @@
...
+    val operation = (request.body \ "operationName").asOpt[String]
+
+    val variables = (request.body \ "variables").toOption.flatMap {
+      case JsString(vars) => Some(parseVariables(vars))
+      case obj: JsObject => Some(obj)
+      case _ => None
+    }
+
+    executeQuery(query, request, variables, operation)
+      .map(_.withHeaders(("Access-Control-Allow-Origin", "*")))
```
bryan-coursera: we can remove the CORS header here

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/GraphQLController.scala
```diff
@@ -0,0 +1,92 @@
...
+      requestHeader: RequestHeader,
+      variables: Option[JsObject],
+      operation: Option[String]) = {
+
+    // TODO(bryan): Handle errors / failures properly here
+    (for {
+      request <- SangriaGraphQlParser.parse(query, requestHeader)
+      document <- QueryParser.parse(query).toOption
+    } yield {
+      val fetcherExecution: Future[Option[Response]] = if (query.contains("IntrospectionQuery")) {
```
bryan-coursera: we should add a todo here to just have the engine handle introspection queries naturally instead of having to special case them here. (in those cases, there are just no data needs, so it should return instantly with an empty map of data)

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/GraphQLController.scala
```diff
@@ -0,0 +1,92 @@
...
+        Future.successful(None)
+      } else {
+        engine.execute(request).map(Some(_))
+      }
+      fetcherExecution.flatMap { responseOpt =>
+        val allData = responseOpt.map(
+          _.data.map { case (resourceName, response) =>
+            s"${resourceName.topLevelName}.v${resourceName.version}" -> response.values.toList
```
bryan-coursera: resourceName has a method to generate the string value... we should use that here. (and yes, I realize I'm probably just reviewing a lot of the code that I wrote. don't judge)

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/GraphqlSchemaProvider.scala
```diff
@@ -0,0 +1,105 @@
...
+    if (numRetries == 0) {
+      logger.warn(s"Could not compute a consistent snapshot of the schema provider.")
+      (schemaProvider.resourceSchemas, schemaProvider.types)
+    } else {
+      val resourceSchemas = schemaProvider.resourceSchemas
+      val types = schemaProvider.types
+      if ((schemaProvider.resourceSchemas eq resourceSchemas) && (schemaProvider.types eq types)) {
+        (resourceSchemas, types)
+      } else {
+        consistentSnapshot(numRetries - 1)
```
bryan-coursera: do we need to set any timeout here, or are you just assuming that the time to consistency will be so small that simply running it again will result in a consistent view? this seems a little funny (and possibly over-engineering.)

is the concern that we might set resourceSchemas, then read resourceSchemas and typesSet, and then set typesSet? If so, maybe we just store them all together in a case class or something that'll reduce code complexity, even if its a bit less performant.

let's chat more tomorrow

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/GraphqlSchemaProvider.scala
```diff
@@ -0,0 +1,105 @@
...
+        builder.generateSchema().asInstanceOf[Schema[SangriaGraphQlContext, Any]]
+      this.resourceSchemas = resourceSchemas
+      this.typesSet = typesSet
+      this.cachedSchema = graphQlSchema
+    } catch {
+      case NonFatal(e) =>
+        logger.error(s"Could not build schema.", e)
+        this.resourceSchemas = resourceSchemas
+        this.typesSet = typesSet
+        this.cachedSchema = DefaultGraphqlSchemaProvider.EMPTY_SCHEMA
```
bryan-coursera: on a failure, do we want to fall back to the empty schema, or instead just keep the existing schema?

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/GraphQLController.scala
saeta: How about this goes into a github issue instead?

saeta:
@bryan-coursera PTAL?

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/GraphQLController.scala
bryan-coursera: ping

naptime/src/main/scala/org/coursera/naptime/ari/models.scala
```diff
@@ -43,17 +60,19 @@ trait FetcherApi {
...
    * @return The merged type schema corresponding to the merged type name.
    */
   def mergedType(resourceName: ResourceName): Option[RecordDataSchema]
+
+  /**
+   * The collection of all Naptime resources available.
+   *
+   * @return a tuple of (1) a set of naptime resource schema and (2) a set of DataSchemas corresponding to known types
+   */
+  def fullSchema: (Set[Resource], Set[DataSchema])
```
bryan-coursera: thoughts on putting this in a case class instead of a tuple?

saeta:
Nit fixes.

PR #110, opened 2016-09-22
Add relay schema support

bryan-coursera:
- Adds connection and edge / node support to schemas
- Update resolvers to parse through edges + nodes
- Modify SangriaGraphQlContext to take in a full response object
- Use response information when pulling lists of models from a request
- Completely omit resources from a schema that don’t have valid handlers or models
- Reduce many errors to warnings
- Fix a couple of problems with the example app
- Bump version number

PTAL @saeta


naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/SangriaGraphQlSchemaBuilder.scala
```diff
@@ -147,6 +176,8 @@ class SangriaGraphQlSchemaBuilder(
     import sangria.marshalling.FromInput.coercedScalaInput
 
     val listPattern = "(set|list|seq|immutable.Seq)\\[(.*)\\]".r
+    val optionPattern = "(Option)\\[(.*)\\]".r
+
```
saeta: Can you please de-dupe with the regex's above?

saeta:
:shipit:

PR #111, opened 2016-09-28
Fix fragment parsing and __id generation

bryan-coursera:
- Fragments on top-level resources weren’t being parsed correctly. Now they are
- global id generation for non-string ids was also broken. now its fixed

PTAL @saeta


saeta:
Additionally, can you add tests to test fragments on further nested models?

bryan-coursera:
@saeta -- I added tests for malformed and deeply nested fragments, but I didn't dig too much on many ways that queries can be malformed. Before we even call the sangria parser, we run the query through the Sangria AST parser, which will pick up on any obvious syntax errors / unparseable queries. Let me know if there are other tests that you had in mind

saeta:
LGTM. Thanks. Once Travis approves, I'll merge this in and publish. Thanks!!

PR #115, opened 2016-10-04
Add DataMap support to JSON serializer and schemas

bryan-coursera:
Some of our Courier models support a `passthroughExempt` flag, which lets the model support an arbitrary, untyped DataMap as a field. GraphQL/Sangria was freaking out when seeing this though, as there were no fields defined (which is not allowed according to GraphQL’s spec). This takes a lot of inspiration from both https://github.com/taion/graphql-type-json and https://gist.github.com/OlegIlyenko/5b96f4b54f656aac226d3c4bc33fd2a6 to add a `DataMapType` to the GraphQL schema we support.

TODO: add tests for this :). But initial thoughts would be nice, in case I have to end up refactoring any of this.

PTAL @saeta , and cc @lewisf as you need this for Nostos support to work properly.

### DIFF
--- a/examples/src/main/pegasus/org/coursera/example/AnyData.courier
+++ b/examples/src/main/pegasus/org/coursera/example/AnyData.courier
@@ -1,0 +1,13 @@
+namespace org.coursera.example
+
+/**
+ * Marker type for arbitrary JSON data.
+ *
+ * WARN: Provided for compatibility with existing data models only. Please define schemas for all
+ * data.
+ *
+ * This will be deprecated once `AnyRecord` is available.
+ */
+@passthroughExempt
+  record AnyData {
+}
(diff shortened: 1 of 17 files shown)


examples/src/main/pegasus/org/coursera/example/AnyData.courier
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * Marker type for arbitrary JSON data.
+ *
+ * WARN: Provided for compatibility with existing data models only. Please define schemas for all
+ * data.
+ *
+ * This will be deprecated once `AnyRecord` is available.
+ */
+@passthroughExempt
+  record AnyData {
```
saeta: spacing

naptime/src/test/pegasus/org/coursera/naptime/ari/graphql/models/AnyData.courier
```diff
@@ -0,0 +1,13 @@
...
+/**
+ * Marker type for arbitrary JSON data.
+ *
+ * WARN: Provided for compatibility with existing data models only. Please define schemas for all
+ * data.
+ *
+ * This will be deprecated once `AnyRecord` is available.
+ */
+@passthroughExempt
+  record AnyData {
```
saeta: indentation

PR #117, opened 2016-10-04
Add response pagination to ARI Local Mode.

saeta:
ARI local mode should include top-level pagination information.


bryan-coursera:
haha, that seems reasonable enough :)

PR #119, opened 2016-10-09
upgrade naptime to play 2.5

zhaojunz:
I want to build a version of play 2.5 of naptime, otherwise, i can't boot service in infra-services.

Reviewers: saeta

CC:

Test Plan: unit test passing, will test it in infra-services

Auditors:

### DIFF
--- a/naptime/src/main/scala/org/coursera/naptime/actions/RestAction.scala
+++ b/naptime/src/main/scala/org/coursera/naptime/actions/RestAction.scala
@@ -16,6 +16,9 @@
 
 package org.coursera.naptime.actions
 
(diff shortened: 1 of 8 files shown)


zhaojunz:
I don't know how to handle this as play 2.5 removes all charset=utf-8 after 2.5. see below for the quote:

> Play WS would infer a charset from the content type and append a charset to the Content-Type header of the request if one was not already set. This caused some confusion and bugs, and so in 2.5.x the Content-Type header does not automatically include an inferred charset. If you explicitly set a Content-Type header, the setting is honored as is.

PR #121, opened 2016-10-12
Add support for variables in GraphQL

bryan-coursera:
Also fixes a few query parameter problems with the LocalFetcher and a bug with a 404 on gets in the Engine2

PTAL @saeta

### DIFF
--- a/naptime-testing/src/test/scala/org/coursera/naptime/NestedMacroCourierTests.scala
+++ b/naptime-testing/src/test/scala/org/coursera/naptime/NestedMacroCourierTests.scala
@@ -15,8 +15,8 @@ import org.coursera.naptime.router2.NaptimeRoutes
 import org.coursera.naptime.router2.Router
 import org.junit.Test
 import org.scalatest.concurrent.ScalaFutures
+import org.scalatest.exceptions.TestFailedException
 import org.scalatest.junit.AssertionsForJUnit
-import org.scalatest.mock.MockitoSugar
 import play.api.libs.json.JsString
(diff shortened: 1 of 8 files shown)


naptime-testing/src/test/scala/org/coursera/naptime/NestedMacroCourierTests.scala
```diff
@@ -139,17 +139,11 @@ class NestedMacroCourierTests extends AssertionsForJUnit with ScalaFutures {
...
-    assert(1 === response.topLevelResponses.size)
-    assert(1 === response.data.size)
-    assert(response.data.contains(CoursesResource.ID))
-    val coursesResponse = response.data(CoursesResource.ID)
-    assert(1 === coursesResponse.size)
-    assert(coursesResponse.contains("abc"))
-    assert("course-abc" === coursesResponse("abc").get("name"),
-      s"Data map: ${coursesResponse("abc")}")
-    // TODO: Check pagination
+    assert(interceptedException.getMessage.contains("Gets are not supported in the LocalFetcher"))
```
saeta: Can you keep this test, but also add a multi-get one?

bryan-coursera: you mean like the one right below?

saeta: Good point. Didn't see that one there. #shipit

PR #122, opened 2016-10-14
Add schema inference override registry.

saeta:
When computing schemas for resources, if the types are neither
courier-based nor primitives, a reflection-based schema-inference
algorithm is run. This works fine for case-classes, but fails miserably
on some of their fields (e.g. if a field is a java.util.Date). This
results in relatively useless schemas (e.g. for GraphQL).

In this commit, I have added a Guice Multi-binding/MapBinding-based
implementation to register these overrides. Additionally, I've added a
utility that can fix up these inferred schemas at runtime based on the
configured overrides. Finally, I've added a test that verifies the
correctness of all of the above.


bryan-coursera:
LGTM.

PR #123, opened 2016-10-14
Major refactor of schema generation

bryan-coursera:
- Removes Relay schemas
- Refactors the schema building stuff to remove duplication and separate different parts of the builder
- Fixes various bugs (especially around multiple top-level requests)

PTAL @saeta


naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/FieldBuilder.scala
```diff
@@ -0,0 +1,167 @@
...
+      case (None, recordDataSchema: RecordDataSchema) =>
+        NaptimeRecordField.build(schemaMetadata, recordDataSchema, fieldName, namespace)
+
+      case (None, enumDataSchema: EnumDataSchema) =>
+        NaptimeEnumField.build(enumDataSchema, fieldName)
+
+      case (None, unionDataSchema: UnionDataSchema) =>
+        NaptimeUnionField.build(schemaMetadata, unionDataSchema, fieldName, namespace)
+
+      case (None, typerefDataSchema: TyperefDataSchema) =>
```
saeta: For typeref data schema, I think you should call `getDereferencedDataSchema` instead of `getRef`. Have you been able to test this code path? (esp. with typeref's of typerefs.)

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeUnionField.scala
```diff
@@ -0,0 +1,67 @@
...
+        resolve = subTypeField.resolve)
+      ObjectType[SangriaGraphQlContext, DataMap](
+        name = FieldBuilder.formatName(s"${subType.getUnionMemberKey}Member"),
+        fields = List(field))
+    }.toList
+    val unionName = buildFullyQualifiedName(namespace.getOrElse(""), fieldName)
+    new UnionType(unionName, None, objects) {
+      // write a custom type mapper to use field names to determine the union member type
+      override def typeOf[Ctx](value: Any, schema: Schema[Ctx, _]): Option[ObjectType[Ctx, _]] =
+      {
```
saeta: formatting nit.

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/FieldBuilder.scala
```diff
@@ -0,0 +1,167 @@
...
+      field: RecordDataSchemaField,
+      namespace: Option[String],
+      fieldNameOverride: Option[String] = None): Field[SangriaGraphQlContext, DataMap] = {
+    type ResolverType = Context[SangriaGraphQlContext, DataMap] => Value[SangriaGraphQlContext, Any]
+
+    val relatedResourceOption = field.getProperties.asScala.get("related").map(_.toString)
+
+    val fieldName = fieldNameOverride.getOrElse(field.getName)
+
+    val fieldDoc = if (field.getDoc.isEmpty) None else Some(field.getDoc)
```
bryan-coursera: I'm not sure I understand. `Field.getDoc` will return the doc from the record type, i.e.
```
record Course {
  /**
   * This field denotes the main instructors for the course
   */
  primaryInstructors: InstructorList

 /**
   * This field denotes the backup instructors for the course
   */
  backupInstructors: InstructorList
}```

We'd want to differentiate in the doc here between primaryInstructors and backupInstructors -- if we went to the documentation on InstructorList, it would be the same for both.

In GraphiQl, people can get the documentation for the field, and then on the type itself, see the documentation. (i.e. documentation on both primaryInstructors and InstructorList would be available).

saeta: Ah, okay, great.

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/FieldBuilder.scala
bryan-coursera: Ah ok. I can make this change, but I think in the way that I'm calling buildField with the ref now, it'll also recursively get to the right type. So both ways should work, but yours might be a bit more efficient. Lemme just double check that things continue to work with both

saeta: Yeah, I agree that as-is should work (because you're doing it recursively).

PR #124, opened 2016-10-17
Handle NullDataSchemas and log error for unknown field types

bryan-coursera:
PTAL @saeta .
0.3.11 currently doesn't generate schemas correctly for our production apis :(


saeta:
Please add a unit test in a subsequent PR to capture the failure, to ensure no regressions in the future.

PR #125, opened 2016-10-18
Add fields for pagination

bryan-coursera:
- Adds a Pagination field on both top-level resource handlers (finders, getAlls, multiGets), as well as for inner related resources
- Bumps version up to 0.4.0

PTAL @saeta . working on tests now, but initial thoughts would be nice. (also want to make sure that this isn't blocking anyone)

### DIFF
--- a/examples/src/main/scala/resources/CoursesResource.scala
+++ b/examples/src/main/scala/resources/CoursesResource.scala
@@ -29,15 +29,16 @@ class CoursesResource @Inject() (
       .map { case (id, course) => Keyed(id, course) }.toList)
   }
 
-  def getAll(limit: Int = 10, start: Option[String]) = Nap.getAll { context =>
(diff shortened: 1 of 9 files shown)


examples/src/main/scala/resources/CoursesResource.scala
```diff
@@ -29,8 +29,18 @@ class CoursesResource @Inject() (
       .map { case (id, course) => Keyed(id, course) }.toList)
   }
 
-  def getAll() = Nap.getAll { context =>
-    Ok(courseStore.all().map { case (id, course) => Keyed(id, course) }.toList)
+  def getAll(limit: Int = 10, start: Option[String]) = Nap.getAll { context =>
```
saeta: Usually, pagination information is handled on the `context` object.

PR #126, opened 2016-10-20
Set `gets` as nullable to handle 404s

bryan-coursera:
Also pulls ParentContext out to its own file just to make things nicer, and creates a new exception class to use.

PTAL @saeta

### DIFF
--- a/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/exceptions.scala
+++ b/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/exceptions.scala
@@ -1,6 +1,9 @@
 package org.coursera.naptime.ari.graphql.schema
 
+import sangria.execution.UserFacingError
+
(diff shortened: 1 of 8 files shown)


naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/exceptions.scala
```diff
@@ -1,4 +1,8 @@
 package org.coursera.naptime.ari.graphql.schema
 
+import sangria.execution.UserFacingError
+
 case class SchemaGenerationException(msg: String) extends Exception(msg)
 case class SchemaExecutionException(msg: String) extends Exception(msg)
+
+case class NotFoundException(msg: String) extends Exception(msg) with UserFacingError
```
saeta: Can this be documented as to what not-found corresponds to? Does it correspond to resource not found? Or element not found? Or something else? Alternatively, rename to something less ambiguous. (i.e. `ElementNotFoundException`.)

PR #129, opened 2016-10-22
Adds a naive approach to calculating complexity

bryan-coursera:
Basic approach is that each field selection counts as 1, remote `get` requests count as 10, and paginated remote requests count as 10 * (limit / 10). I imagine we’ll be tweaking this _a lot_ as we go, but this will at least prevent against a lot of basic attacks possible on the graphql endpoint.

Thoughts, @saeta ? and maybe @jnwng and @lewisf have thoughts too

### DIFF
--- a/naptime-graphql/src/test/scala/org/coursera/naptime/ari/graphql/schema/NaptimePaginatedResourceFieldTest.scala
+++ b/naptime-graphql/src/test/scala/org/coursera/naptime/ari/graphql/schema/NaptimePaginatedResourceFieldTest.scala
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2016 Coursera Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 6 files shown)


naptime-graphql/src/test/scala/org/coursera/naptime/ari/graphql/schema/NaptimePaginatedResourceFieldTest.scala
```diff
@@ -0,0 +1,57 @@
...
+    assert(limitFifty === 5 * NaptimePaginatedResourceField.COMPLEXITY_COST * 1)
+
+    val limitZero = field.complexity.get.apply(context, Args(Map("limit" -> 0)), 1)
+    assert(limitZero === 1 * NaptimePaginatedResourceField.COMPLEXITY_COST * 1)
+
+    val childScoreFive = field.complexity.get.apply(context, Args(Map("limit" -> 1)), 5)
+    assert(childScoreFive === 1 * NaptimePaginatedResourceField.COMPLEXITY_COST * 5)
+
+  }
+
```
saeta: Could you add one test that goes from GraphQL query text to a complexity number?

PR #130, opened 2016-10-22
Add engine metrics collector to the Engine

bryan-coursera:
- provides a trait that can be overridden by clients to collect and report metrics from the engine into a downstream system (i.e. DataDog)
- Returns the number of downstream requests from the engine as a response header on GraphQL requests (for developer debugging purposes)

@saeta -- thoughts on this approach? I'm working on tests for this now, but want to get your opinion on things before I spend too much time on tests (in case I have to redo some of this)

### DIFF
--- a/naptime/src/main/scala/org/coursera/naptime/ari/models.scala
+++ b/naptime/src/main/scala/org/coursera/naptime/ari/models.scala
@@ -126,7 +126,14 @@ case class RequestField(
 
 case class ResponseMetrics(
   numRequests: Int = 0,
-  duration: FiniteDuration = FiniteDuration(0, "seconds"))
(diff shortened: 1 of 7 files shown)


naptime/src/main/scala/org/coursera/naptime/ari/models.scala
```diff
@@ -123,6 +124,10 @@ case class RequestField(
   args: Set[(String, JsValue)], // TODO: Should JsValue be a specific ARI type?
   selections: List[RequestField])
 
+case class ResponseMetrics(
+  numRequests: Int = 0,
+  duration: FiniteDuration = FiniteDuration(0, "seconds"))
```
saeta: define a `++` operator here.

PR #134, opened 2016-10-29
Support recursive models in fixupInferredSchemas

bryan-coursera:
The schema fixup helper recursively traverses through recorddataschemas. However, some schemas are recursive, which would lead to a StackOverthrowException here. To fix it, we’ll keep track of the nodes that we’ve visited to ensure that we don’t get stuck in a loop

PTAL @saeta or @yifan-coursera


saeta:
LGTM.

PR #135, opened 2016-10-30
Default the type for a key field to string

bryan-coursera:
There are some key types that are non-courier’d that can’t have their types inferred through reflection. This can occur when the type is or includes a Java class inside of it (i.e. java.util.UUID). In these cases, we can assume that there is a StringKeyFormat available to convert the key type to a string, so it is safe to assume that the ID in the response will be of type string.

In these cases, when the key type cannot be inferred, it will fall back to a string. By falling back, we can ensure that the merged type can be generated correctly, and add graphql support to these resources.


saeta:
Note: test failure looks like a future timeout.

PR #136, opened 2016-10-31
Properly handle complexity errors and mark SchemaProvider as a Singleton

bryan-coursera:
The complexity measurement can also throw query validation errors, so this properly catches them too. Also marks DefaultGraphqlSchemaProvider as a Singleton to ensure it boots up quickly

PTAL @yifan-coursera / cc @saeta


yifan-coursera:
LGTM

bryan-coursera:
going to close this, because https://github.com/coursera/naptime/pull/137 handles this same case, but better

PR #137, opened 2016-11-01
Add filter support to GraphQL Controller

bryan-coursera:
Pulls out some non-controller stuff into a series of filters. This will also allow naptime clients to specify their own filters, which can be used for more implementation-specific things (i.e. reporting to Datadog), implementing better security restrictions, etc.

This design was modeled heavily after Play's filter approach: https://www.playframework.com/documentation/2.5.x/ScalaHttpFilters


naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/filters/QueryComplexityFilter.scala
```diff
@@ -0,0 +1,79 @@
...
+        OutgoingQuery(InternalServerError(Json.obj("error" -> error.getMessage)), None)
+    }
+  }
+
+  private[graphql] def computeComplexity(
+      queryAst: Document,
+      variables: JsObject): Future[Double] = {
+    // TODO(bryan): is there a way around this var?
+    var complexity = 0D
+    val complReducer = QueryReducer.measureComplexity[SangriaGraphQlContext] { (c, ctx) ⇒
```
yifan-coursera: Replace `⇒` with `=>`

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/filters/QueryComplexityFilter.scala
```diff
@@ -0,0 +1,79 @@
...
+      complexity = c
+      ctx
+    }
+    val executorFut = Executor.execute(
```
yifan-coursera: Don't understand the purpose of this since we don't use the result. Is it to validate the query?

bryan-coursera: so because i put a queryReducer here, it'll run the code above on every single node:
```lang=scala
QueryReducer.measureComplexity[SangriaGraphQlContext] { (c, ctx) =>
  complexity = c
  ctx
}
```

That updates a var `complexity`, which we then return later on.

yifan-coursera:
LGTM

PR #138, opened 2016-11-02
Add DataSchemas for parameter types on naptime handlers

bryan-coursera:
This change adds a new `typeSchema` field to the Parameter, which contains a DataSchema (serialized as a DataMap) that defines the parameter type. This allows the type of a parameter to be serialized in a better way than a string, so it can be parsed by the GraphQL assembler and used in schema generation. (Presently, the scala typename is being parsed with a regex, but that’s terrible)

Once this is landed, the schema store will be updated to parse the correct parameter types, and then the GraphQL schema builder can use that type to generate the graphql parameters.

### DIFF
--- a/naptime-models/src/main/scala/org/coursera/naptime/courier/SchemaInference.scala
+++ b/naptime-models/src/main/scala/org/coursera/naptime/courier/SchemaInference.scala
@@ -48,6 +48,9 @@ object SchemaInference {
    */
   def inferSchema[T: ru.TypeTag]: JsObject = inferSchema(ru.typeOf[T])
 
+  def inferSchemaFromWeakTypeTag[T: ru.WeakTypeTag]: JsObject = inferSchema(ru.weakTypeOf[T])
+
+
   /**
    * Extracts schemas from Courier generated bindings.
    * Infers schemas from Play! JSON style classes that use `Format` and `OFormat`.
--- a/naptime/src/main/scala/org/coursera/naptime/router2/MacroImpls.scala
+++ b/naptime/src/main/scala/org/coursera/naptime/router2/MacroImpls.scala
@@ -16,9 +16,6 @@
 
 package org.coursera.naptime.router2
 
-import com.linkedin.data.schema.IntegerDataSchema
-import com.linkedin.data.schema.LongDataSchema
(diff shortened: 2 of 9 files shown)


naptime/src/main/scala/org/coursera/naptime/router2/MacroImpls.scala
```diff
@@ -243,6 +243,33 @@ class MacroImpls(val c: blackbox.Context) {
       }
     }
 
+    private[this] def getDataSchemaDataMapForType(targetType: c.Type): c.Tree = {
+      val schema = if (targetType <:< SCALA_RECORD_TEMPLATE) {
```
yifan-coursera: Why not get this from `getRecordSchemaForType`? (Macro weirdness?)

naptime-models/src/main/scala/org/coursera/naptime/courier/SchemaInference.scala
```diff
@@ -48,6 +48,9 @@ object SchemaInference {
    */
   def inferSchema[T: ru.TypeTag]: JsObject = inferSchema(ru.typeOf[T])
 
+  def inferSchemaFromWeakTypeTag[T: ru.WeakTypeTag]: JsObject = inferSchema(ru.weakTypeOf[T])
+
+
```
yifan-coursera: Delete extra newline.

naptime/src/test/scala/org/coursera/naptime/actions/RecordTemplateNaptimeSerializerTest.scala
```diff
@@ -30,12 +30,13 @@ class RecordTemplateNaptimeSerializerTest extends AssertionsForJUnit {
...
   @Test
   def simpleTest(): Unit = {
-    val parameter = Parameter("parameterName", "fakeType", List.empty)
+    val parameter = Parameter("parameterName", "fakeType", None, List.empty)
 
     val expected = new DataMap()
     expected.put("name", "parameterName")
     expected.put("type", "fakeType")
     expected.put("attributes", new DataList())
+    expected.put("required", Boolean.box(false))
```
yifan-coursera: Whoa, I haven't seen an explicit box in forever.

PR #139, opened 2016-11-07
Fix bug with field names for enum fields

bryan-coursera:
Minor fix for field names on enumfields.

PTAL @yifan-coursera


yifan-coursera:
LGTM.

PR #140, opened 2016-11-07
Fix bug with variables in query execution

bryan-coursera:
Self explanatory. Must've been from the refactor.

PTAL @yifan-coursera


yifan-coursera:
LGTM

PR #141, opened 2016-11-10
Support query batching in the naptime controller

bryan-coursera:
Note: this changes the way that the filters work to instead pass around JsObjects instead of a response. This means that filters / the GraphQL engine cannot alter the response at all (including response codes + headers), but instead must put all data on the response body and _always_ return a 200 access code. This is acceptable with the way that GraphQL works, but is a change to keep in mind

### DIFF
--- a/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/GraphQLController.scala
+++ b/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/GraphQLController.scala
@@ -62,6 +62,7 @@ class GraphQLController @Inject() (
   }
 
   def graphqlBody = Action.async(parse.json) { request =>
+
     val query = (request.body \ "query").as[String]
     val operation = (request.body \ "operationName").asOpt[String]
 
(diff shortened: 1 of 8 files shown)


yifan-coursera:
LGTM

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/GraphQLController.scala
```diff
@@ -71,7 +72,29 @@ class GraphQLController @Inject() (
...
+      val variables = (queryObj \ "variables").toOption.flatMap {
+        case JsString(vars) => Some(parseVariables(vars))
+        case obj: JsObject => Some(obj)
+        case _ => None
+      }.getOrElse(Json.obj())
+
+      executeQuery(query, request, variables, operation)
+    }
+    resultsFut.map { results =>
+      Ok(Json.arr(results.map(_.response)))
```
jnwng: something is odd about what's being returned here. image

unfortunately im not certain what's going on here other than there's too many maps or something here

PR #143, opened 2016-11-16
Fix bugs with generating and resolving union types

bryan-coursera:
This fixes two bugs with union types, both around the way that names are converted from a courier name (i.e. org.coursera.UnionMember) to a GraphQL compatible name (org_coursera_UnionMember).

The first bug was that we could not correctly resolve the ObjectType to parse into based on the field names, because we were comparing strings of the two different formats, which were never matching. This wasn't caught when we were working with primitive types, because "int" didn't have any field name conversion issues.

The second bug was around how we were parsing the data out of the record field, which again was due to a mismatch in the two names.

These issues should be fixed now, and I added a unit test to check against more complex union types.


naptime-graphql/src/test/scala/org/coursera/naptime/ari/graphql/DefaultGraphqlSchemaProviderTest.scala
```diff
@@ -92,6 +92,11 @@ object DefaultGraphqlSchemaProviderTest extends MockitoSugar {
     "org_coursera_naptime_ari_graphql_models_Coordinates",
     "org_coursera_naptime_ari_graphql_models_CoursePlatform",
     "org_coursera_naptime_ari_graphql_models_originalId",
+    "org_coursera_naptime_ari_graphql_models_OldPlatformData",
+    "org_coursera_naptime_ari_graphql_models_platformSpecificData",
+    "org_coursera_naptime_ari_graphql_models_NewPlatformDataMember",
+    "org_coursera_naptime_ari_graphql_models_NewPlatformData",
+    "org_coursera_naptime_ari_graphql_models_OldPlatformDataMember",
```
yifan-coursera: Nit: put this right after `OldPlatformData` for tidiness

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeUnionField.scala
```diff
@@ -36,10 +36,10 @@ object NaptimeUnionField {
         schemaMetadata,
         new RecordDataSchemaField(subType),
         namespace,
-        Some(fieldName))
+        Some(subType.getUnionMemberKey))
 
       val field = Field.apply[SangriaGraphQlContext, DataMap, Any, Any](
-        FieldBuilder.formatName(subType.getUnionMemberKey),
+        fieldName,
```
yifan-coursera: q: Why doesn't this need to be formatted?

bryan-coursera: fieldName is the formatted name already (line 34)

PR #145, opened 2016-11-16
Fix a typo and prevent double-nesting of batched responses

bryan-coursera:
PTAL @jnwng


jnwng:
let's go to master

bryan-coursera:
going to master...

PR #147, opened 2016-11-27
Add support for reverse relations to the ARI engine

bryan-coursera:
While forward relations are very useful in a majority of cases (i.e. courses -> instructors, where the course model has an instructorId on it), there are some cases where a relation is necessary where the first model doesn't have any reference to the second. For instance, while a Membership model has a CourseId available on it, the Course model has no reference to all of the memberships available.

In order to create this course -> membership relationship though, we can dynamically fetch the relationship via a query to the membership resource, which, given information about the course, can return a list of memberships + ids for the relationship. This is referred to as a "reverse relationship", where the relationship data is stored on the related model.

These relationships are defined in the Naptime resource (via the `Fields` attribute), like so:
```lang=scala
override implicit lazy val val Fields = Fields
  .withReverseRelations(
    "memberships" -> ReverseRelation[MembershipId](
      resourceName = ResourceNames.membershipResourceName,
      arguments = Map(
        "q" -> "meByCourse",
        "courseId" -> "$id",
        "showHidden" -> "true")))
```

The `$id` string is treated like string interpolation, and will be replaced with the value of the "id" field on the fetched course.

During execution, the engine will make the API call as described on the relationship (appending all interpolated arguments listed), and will make that response available. It will also create a new dynamic field on the original model (in this case, Course), with the ids of the related resources for easier traversal / interop with forward relations.

PTAL @yifan-coursera  -- let me know if you want to go over this one in person.

And if you're still interested, cc @saeta

### DIFF
--- a/examples/src/main/pegasus/org/coursera/example/Course.courier
+++ b/examples/src/main/pegasus/org/coursera/example/Course.courier
@@ -13,4 +13,6 @@ record Course {
   description: string?
 
   extraData: AnyData
+
+  courseMetadata: union[CourseMetadata]
 }
--- a/naptime/src/main/scala/org/coursera/naptime/Types.scala
+++ b/naptime/src/main/scala/org/coursera/naptime/Types.scala
@@ -111,7 +111,7 @@ object Types extends StrictLogging {
           newField.setRecord(mergedSchema)
 
           val reverseRelatedMap = Map[String, AnyRef](
-            Relations.REVERSE_PROPERTY_NAME -> reverseRelation.toAnnotation().data())
+            Relations.REVERSE_PROPERTY_NAME -> reverseRelation.toAnnotation.data())
           newField.setProperties(reverseRelatedMap.asJava)
 
           val existingFields = mergedSchema.getFields
--- a/naptime/src/main/scala/org/coursera/naptime/Types.scala
+++ b/naptime/src/main/scala/org/coursera/naptime/Types.scala
@@ -18,6 +18,7 @@ package org.coursera.naptime
 
 import java.lang.StringBuilder
 
+import com.linkedin.data.schema.ArrayDataSchema
 import com.linkedin.data.schema.DataSchema
 import com.linkedin.data.schema.Name
 import com.linkedin.data.schema.PrimitiveDataSchema
@@ -26,12 +27,14 @@ import com.linkedin.data.schema.RecordDataSchema.RecordType
 import com.linkedin.data.schema.StringDataSchema
 import com.linkedin.data.schema.TyperefDataSchema
 import com.typesafe.scalalogging.StrictLogging
+
 import scala.collection.JavaConverters._
 
 object Types extends StrictLogging {
 
   object Relations {
     val PROPERTY_NAME = "related"
+    val REVERSE_PROPERTY_NAME = "relatedOn"
   }
 
   @deprecated("Please use the one with fields included", "0.2.4")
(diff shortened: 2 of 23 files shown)


saeta:
Wahoo!

naptime/src/main/scala/org/coursera/naptime/ari/engine/EngineHelpers.scala
```diff
@@ -0,0 +1,228 @@
...
+      schema: RecordDataSchema,
+      path: Seq[String],
+      data: AnyRef): Unit = {
+    val it = Builder
+      .create(element, schema, IterationOrder.PRE_ORDER)
+      .dataIterator()
+    var dataElement: DataElement = null
+    while ( {
+      dataElement = it.next(); dataElement
+    } != null) {
```
yifan-coursera: More idiomatic to write this as 
```
Iterator
  .continually(it.next)
  .takeWhile(_ != null)
  .filter(_.path.toSeq.map(_.toString) == path.dropRight(1))
  .foreach(_.getValue.asInstanceOf[DataMap].put(path.last, data)
```

naptime/src/main/scala/org/coursera/naptime/ari/engine/EngineHelpers.scala
```diff
@@ -0,0 +1,228 @@
...
+      data: Iterable[DataMap],
+      schema: RecordDataSchema): (List[ForwardRelatedField], List[ReverseRelatedField]) = {
+
+    val forwardRelations = mutable.Buffer[ForwardRelatedField]()
+    val reverseRelations = mutable.Buffer[ReverseRelatedField]()
+
+    data.foreach { element =>
+      val it = Builder.create(element, schema, IterationOrder.PRE_ORDER).dataIterator()
+      var dataElement: DataElement = null
+      while ( {
```
yifan-coursera: Do an iterator here too, as above.

examples/src/main/pegasus/org/coursera/example/Course.courier
```diff
@@ -13,4 +13,6 @@ record Course {
   description: string?
 
   extraData: AnyData
+
+  courseMetadata: union[CourseMetadata]
```
bryan-coursera: just for testing that relations on unions work. i can remove this field if you want

naptime/src/main/scala/org/coursera/naptime/ari/engine/EngineHelpers.scala
bryan-coursera: oh woah. i hadn't used scala iterators before... that's really nice though :)

naptime/src/main/scala/org/coursera/naptime/Types.scala
```diff
@@ -95,6 +98,27 @@ object Types extends StrictLogging {
...
       }
     }
+    for ((name, reverseRelation) <- fields.reverseRelations) {
+      Option(mergedSchema.getField(name)) match {
+        case Some(field) =>
+          logger.warn(s"Fields for resource $typeName tries to add reverse relation on field " +
+            s"'$name', but that field is already defined on the model")
+        case None =>
+          val errorMessageBuilder = new StringBuilder
+          val newField = new RecordDataSchema.Field(new ArrayDataSchema(new StringDataSchema)) // TODO(bryan): fix type here
```
bryan-coursera: i believe it still works, because ids are serialized to strings in naptime output. I need to investigate this more though and try cases with integers and stuff. this is why i put the key type on the reverse relation annotation (`FinderReverseRelation[KeyType](...)`), but i don't know if that's necessary or if i can always assume the id will be a string.

i'm going to keep this todo in here and investigate when there's more time, but i don't think it'll break anything now.

PR #151, opened 2016-11-30
Add a placeholder value for empty enums to support GraphQL syntax

bryan-coursera:
GraphiQL doesn’t like when there’s an enum defined without any member values. However, that is valid syntax in Courier. In order to handle this case and prevent graphiql from breaking in the future, we’ll simply add an “$UNKNOWN” member value to all empty enums.

PTAL @jnwng and @yifan-coursera


jnwng:
lgtm

PR #152, opened 2016-11-30
Pull in arguments from multiget on singular related field

bryan-coursera:
i.e. a Membership -> Course should still expose the `showHidden` parameter on the courses.v1 multiget

PTAL @yifan-coursera and @jnwng


jnwng:
lgtm

PR #154, opened 2016-12-01
Fix `start` pagination for non-string fields (i.e. ints)

bryan-coursera:
When the id field is a non-string value, pagination using `start` was broken because it was comparing an int and start value. This just calls toString on the id value to always compare strings

PTAL @jnwng / cc @yifan-coursera


jnwng:
lgtm

PR #160, opened 2016-12-11
Support defining relations on nested models / unions

bryan-coursera:
This change allows a relation (either forward or reverse), defined on the `Fields` object in a resource, to reference deeply nested records and unions.

For example, if a Course model has a union of Instructors and TeachingAssistants, a related resource can be added to the Instructor model through the union.

This also changes the variable interpolation to support referencing these nested models too.

### DIFF
--- a/naptime/src/main/scala/org/coursera/naptime/Types.scala
+++ b/naptime/src/main/scala/org/coursera/naptime/Types.scala
@@ -201,6 +201,10 @@ object Types extends StrictLogging {
           val newFields = (existingFields.asScala ++ List(field)).asJava
           val errorMessageBuilder = new StringBuilder
           recordDataSchema.setFields(newFields, errorMessageBuilder)
+          val error = errorMessageBuilder.toString
+          if (error.nonEmpty) {
+            logger.warn("Error while inserting field", error)
+          }
           recordDataSchema
         } else {
           val fieldOption = Option(recordDataSchema.getField(location.head))
--- a/naptime/src/main/scala/org/coursera/naptime/Types.scala
+++ b/naptime/src/main/scala/org/coursera/naptime/Types.scala
@@ -20,12 +20,14 @@ import java.lang.StringBuilder
 
 import com.linkedin.data.schema.ArrayDataSchema
 import com.linkedin.data.schema.DataSchema
+import com.linkedin.data.schema.MapDataSchema
 import com.linkedin.data.schema.Name
 import com.linkedin.data.schema.PrimitiveDataSchema
 import com.linkedin.data.schema.RecordDataSchema
 import com.linkedin.data.schema.RecordDataSchema.RecordType
 import com.linkedin.data.schema.StringDataSchema
 import com.linkedin.data.schema.TyperefDataSchema
+import com.linkedin.data.schema.UnionDataSchema
 import com.typesafe.scalalogging.StrictLogging
 import org.coursera.naptime.schema.RelationType.FINDER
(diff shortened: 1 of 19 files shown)


naptime/src/main/scala/org/coursera/naptime/ari/engine/EngineHelpers.scala
```diff
@@ -110,6 +110,29 @@ object EngineHelpers extends StrictLogging {
...
+    */
+  private[engine] def getValueAtPath(
+      element: DataMap,
+      schema: RecordDataSchema,
+      path: Seq[String]): Option[Object] = {
+      val it = Builder.create(element, schema, IterationOrder.PRE_ORDER).dataIterator()
+      Iterator
+        .continually(it.next)
+        .takeWhile(_ != null)
+        .filter(_.path.toSeq.map(_.toString) == path.dropRight(1))
```
yifan-coursera: Would `find` work here?

naptime/src/main/scala/org/coursera/naptime/ari/engine/EngineImpl.scala
```diff
@@ -211,20 +211,23 @@ class EngineImpl @Inject() (
...
+      val InterpolationRegex = new Regex("""\$(?:([a-zA-Z0-9_]+)|\{([^\}]+)\})""", "withoutBraces", "withBraces")
       val arguments: Set[(String, JsValue)] = reverse.arguments
         .mapValues { value =>
           InterpolationRegex.replaceAllIn(
             value, { regexMatch =>
-              val variableName = regexMatch.group("variableName")
-              Option(topLevelElement.get(variableName)).map {
+              val withoutBraces = Option(regexMatch.group("withoutBraces"))
+              val withBraces = Option(regexMatch.group("withBraces"))
+              val variableName = withoutBraces.orElse(withBraces).getOrElse("")
```
yifan-coursera: What's the optional braces for?

naptime/src/main/scala/org/coursera/naptime/Types.scala
```diff
@@ -194,4 +182,48 @@ object Types extends StrictLogging {
...
+      case mapDataSchema: MapDataSchema =>
+        insertFieldAtLocation(mapDataSchema.getValues, location, field)
+      case arrayDataSchema: ArrayDataSchema =>
+        insertFieldAtLocation(arrayDataSchema.getItems, location, field)
+      case recordDataSchema: RecordDataSchema =>
+        if (location.isEmpty) {
+          field.setRecord(recordDataSchema)
+          val existingFields = recordDataSchema.getFields
+          val newFields = (existingFields.asScala ++ List(field)).asJava
+          val errorMessageBuilder = new StringBuilder
```
bryan-coursera: sure, why not

naptime/src/main/scala/org/coursera/naptime/ari/engine/EngineHelpers.scala
bryan-coursera: i believe so. good call

naptime/src/main/scala/org/coursera/naptime/ari/engine/EngineImpl.scala
bryan-coursera: ah. i'll leave a comment. i want to match both:
`"$instructorIds"` and `"${instructorUnion/org.coursera.InstructorUnionType/instructorIds}"`

yifan-coursera:
LGTM

PR #165, opened 2017-01-06
Fix escaping exceptions in StringKey parsing

kyewei:
In the `read` provided by `recordTemplateStringKeyFormat`, if the tuple [here](https://github.com/kyewei/naptime/blob/cb5290356348a1fba99cd850415047298c66698f/naptime-models/src/main/scala/org/coursera/naptime/courier/StringKeyCodec.scala#L198) fails the check, the error leaks out and a `None` is not returned. Similarly, if a validation in `validateAndFixUp()` fails, the exceptions escapes as well.

Changes: Wrapped lines in try-catch and added tests.

@yifan-coursera

### DIFF
--- a/naptime-models/src/test/scala/org/coursera/naptime/courier/CourierFormatsTest.scala
+++ b/naptime-models/src/test/scala/org/coursera/naptime/courier/CourierFormatsTest.scala
@@ -209,6 +209,17 @@ class CourierFormatsTest extends AssertionsForJUnit {
 
     val roundTripped = StringKey.unapply(stringKey).get
     assert(roundTripped === string)
+
+    val string2 = "nottwo"
+    val stringKey2 = StringKey(string2)
+    val mock2 = stringKey2.asOpt[MockRecord]
+
+    val string3 = "value~notnumeric"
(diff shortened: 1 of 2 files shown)


naptime-models/src/test/scala/org/coursera/naptime/courier/CourierFormatsTest.scala
```diff
@@ -209,6 +209,18 @@ class CourierFormatsTest extends AssertionsForJUnit {
...
+    val stringKey2 = StringKey(string2)
+    val mock2 = stringKey2.asOpt[MockRecord]
+
+    val string3 = "value~notnumeric"
+    val stringKey3 = StringKey(string3)
+    val mock3 = stringKey3.asOpt[MockRecord]
+
+    assert(mock2 === None) // IOException
+    assert(mock3 === None) // DataValidationException
+
```
yifan-coursera: Nit: Remove newline.

kyewei: I amended the original commit so commit history is clean.

PR #169, opened 2017-01-19
Allow setting docs on reverse-relation fields

bryan-coursera:
This lets us write documentation describing the "edge" field that is auto-created by a reverse relation. For instance, if we add a new PrimaryPartners field on a Course that is populated by some arbitrary finder, we can describe on the field what that relationship means.


saeta:
Lgtm

PR #171, opened 2017-02-06
Upgrade courier to 2.0.8 , changing `apply` to `build` as necessary.

amory-coursera:


amory-coursera:
@bryan-coursera PTAL.

PR #173, opened 2017-02-26
Use asScala explicitly in MacroImpl

zhaojunz:
Use the explicit collection.JavaConverters instead of implicit conversions in collection.JavaConversions.

JavaConversions is deprecated in 2.12. And due to the macro impl, this breaks linter we plan to add into infra-services.

http://www.scala-lang.org/api/2.12.x/scala/collection/JavaConversions$.html


zhaojunz:
@josh-newman @bryan-coursera ptal thanks!

zhaojunz:
I didn't, but I think it should be safe.

1) I think the unit tests cover this fairly well (i tried to break the code, and a few unit tests were broken). 
2) Per scala document, JavaConverters is the explicit equivalent of JavaConversions

PR #176, opened 2017-03-01
Make StringKey.asOpt with CourierFormats return None instead of throwing exceptions

### LINKED ISSUE #174: StringKey("!~").asOpt[IntArray] throws exception instead of returning None
`StringKey.asOpt` throws an exception instead of returning none if `StringKeyCodec` on input. This breaks the contract that this method should return `None` upon encountering an error.

```
import org.coursera.common.stringkey.StringKey
import org.coursera.naptime.courier.CourierFormats
import org.coursera.courier.data.IntArray

implicit val stringKeyFormat = CourierFormats.arrayTemplateStringKeyFormat[IntArray]
val key = StringKey("!~").asOpt[IntArray]
```

Expected: 
```None```

Actual:
```
java.io.IOException: `,' expected but ! found line: 1 column: 1
  at org.coursera.naptime.courier.StringKeyCodec$StringListParser.handleParseErrors(StringKeyCodec.scala:382)
  at org.coursera.naptime.courier.StringKeyCodec$StringListParser.parse(StringKeyCodec.scala:353)
  at org.coursera.naptime.courier.StringKeyCodec$Parser.parseList(StringKeyCodec.scala:182)
  at org.coursera.naptime.courier.StringKeyCodec.readList(StringKeyCodec.scala:144)
  at org.coursera.naptime.courier.StringKeyCodec.bytesToList(StringKeyCodec.scala:140)
  at org.coursera.naptime.courier.CourierFormats$$anon$4.reads(CourierFormats.scala:342)
  at org.coursera.common.stringkey.StringKey.asOpt(StringKey.scala:32)
  ... 43 elided
```

### LINKED ISSUE #175: Using StringKey.asOpt and CourierFormats.recordTemplateFormats with records containing typerefs can result in bad lazy vals
Suppose we have a custom class `TestPositiveInt` that represents positive integers and has a coercer to `Int`. 

`TestPositiveInt.scala`:
```
package org.coursera.naptime.courier

import com.linkedin.data.template.Custom
import com.linkedin.data.template.DirectCoercer

case class TestPositiveInt(value: Int) extends AnyVal

object TestPositiveInt {
  object Coercer extends DirectCoercer[TestPositiveInt] {
    override def coerceInput(obj: TestPositiveInt): AnyRef = {
      Int.box(obj.value)
    }

    override def coerceOutput(obj: Any): TestPositiveInt = {
      obj match {
        case value: java.lang.Integer =>
          if (value > 0) {
            TestPositiveInt(value)
          } else {
            throw new IllegalArgumentException(s"$value is not positive")
          }
        case _: Any =>
          throw new IllegalArgumentException()
      }
    }

    def registerCoercer(): Unit = {
      Custom.registerCoercer(new TestPositiveIntCoercer, classOf[TestPositiveInt])
    }
  }

  Coercer.registerCoercer()
}
```

`TestPositiveInt.courier`:
```
namespace org.coursera.naptime.courier

@scala.class = "org.coursera.naptime.courier.TestPositiveInt"
@scala.coercerClass = "org.coursera.naptime.courier.TestPositiveInt.Coercer"

typeref TestPositiveInt = int
```

`TestPositiveIntKeyTest.courier`:
```
namespace org.coursera.naptime.courier

import org.coursera.naptime.courier.TestPositiveInt

record TestPositiveIntKey {
  value: TestPositiveInt
}

```

Now we try to parse a negative integer as this key type:

```
import org.coursera.common.stringkey.StringKey
import org.coursera.naptime.courier.CourierFormats
import org.coursera.naptime.courier.TestPositiveIntKey

implicit val stringKeyFormat = CourierFormats.arrayTemplateStringKeyFormat[TestPositiveIntKey]
val key = StringKey("-1").asOpt[TestPositiveIntKey]
```

Expected:
`key` is `None`

Actual:
`key` is instantiated of type `Some[TestPositiveIntKey]`. It has a `lazy val` called `value`. Upon accessing `value` (such as by printing `key` to the console), `TestPositiveInt.Coercer.coerceOutput` is run and causes an exception to be thrown.

The same error would have occurred if the validation was performed in the constructor of `TestPositiveInt` instead of the coercer.

yifan-coursera:
- Fixes #174: Make `StringKey.asOpt` with `CourierFormats.arrayTemplateStringKeyFormat` return `None` instead of throwing `IOException` on malformed string
- Fixes #175: Make `StringKey.asOpt` with `CourierFormats` return `None` instead of a record with malformed lazy vals if the input would fail validation in the custom Scala coercer or constructor for typeref types


amory-coursera:
Looks good to me. FYI @mkovacs-coursera in case you're still thinking about validation-on-read behavior and this code is useful.

PR #181, opened 2017-04-13
Encode that CourierCollectionResource is top-level

### LINKED ISSUE #168: `documentation` method does not print version on some resources where it should
We have a resource like

```scala
@Singleton
class EnterpriseVoucherResource @Inject() (
    manager: EnterpriseVoucherManager,
    siteAuths: SiteAuths,
    superuserAuths: SuperuserAuths)
    (implicit ec: ExecutionContext)
  extends CourierCollectionResource[EnterpriseVoucherId, EnterpriseVoucher] {

  override def resourceName = "enterpriseVouchers"
  override def resourceVersion = 1
}
```

When printing out the documentation for this resource, I'd expect the paths to look like `enterpriseVouchers.v1/$id`. Instead, they look like `enterpriseVouchers/$id`.

It looks like it must be hitting this branch:
https://github.com/coursera/naptime/blob/master/naptime/src/main/scala/org/coursera/naptime/router2/NaptimePlayRouter.scala#L164

Original conversation at: https://coursera.slack.com/archives/dragon/p1484855939000114

eeasley-coursera:
Should fix #168.

### DIFF
--- a/naptime/src/main/scala/org/coursera/naptime/router2/MacroImpls.scala
+++ b/naptime/src/main/scala/org/coursera/naptime/router2/MacroImpls.scala
@@ -22,6 +22,7 @@ import org.coursera.common.stringkey.StringKeyFormat
 import org.coursera.courier.templates.ScalaRecordTemplate
 import org.coursera.naptime.actions._
 import org.coursera.naptime.resources.CollectionResource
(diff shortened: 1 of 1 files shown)


eeasley-coursera:
@bryan-coursera ?

PR #182, opened 2017-05-04
Support typedDefinitions in unions

bryan-coursera:
Many of our Courier unions use `typedDefinitions`, which allows providing a “friendly name” to a union type, instead of using the fully qualified name.

This change adds supports for these typedDefinitions to Courier, both by changing the name of the member key in the union type, as well as parsing the typedDefinition response correctly.

PTAL @yifan-coursera / cc @qtran-coursera

### DIFF
--- a/naptime-graphql/src/test/scala/org/coursera/naptime/ari/graphql/schema/NaptimeUnionFieldTest.scala
+++ b/naptime-graphql/src/test/scala/org/coursera/naptime/ari/graphql/schema/NaptimeUnionFieldTest.scala
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2016 Coursera Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.coursera.naptime.ari.graphql.schema
+
+import com.linkedin.data.schema.DataSchema
+import com.linkedin.data.schema.IntegerDataSchema
+import com.linkedin.data.schema.Name
+import com.linkedin.data.schema.RecordDataSchema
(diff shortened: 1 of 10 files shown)


naptime-graphql/src/test/scala/org/coursera/naptime/ari/graphql/schema/NaptimeUnionFieldTest.scala
```diff
@@ -0,0 +1,113 @@
...
+      ObjectType("courses_v1_hardMember", List(
+        NaptimeRecordField.build(
+          schemaMetadata,
+          complexFieldDataSchema,
+          "hard",
+          Some("org.coursera.naptime"),
+          resourceName))))
+    val expectedField = UnionType("courses_v1_typedDefinitionTestField", None, expectedUnionTypes)
+    println(field.fieldType.toString)
+    println(expectedField.toString)
```
yifan-coursera: Remove printlns

naptime-graphql/src/test/scala/org/coursera/naptime/ari/graphql/schema/NaptimeUnionFieldTest.scala
```diff
@@ -0,0 +1,113 @@
...
+
+  @Test
+  def build_SingleElementUnion() = {
+    val values = List(new IntegerDataSchema())
+    val union = buildUnionDataSchema(values)
+    val fieldName = "intOnlyUnion"
+    val field = NaptimeUnionField.build(schemaMetadata, union, fieldName, None, resourceName)
+
+    val expectedUnionTypes = List(
+      ObjectType("courses_v1_intMember", List(
```
yifan-coursera: Where does `courses_v1_intMember` come from? I'm not so familiar with how names in unions work.

bryan-coursera: names for unions are a combination of the resource name and the union member name. so this is the wrapper for an int in courses.v1.

(the reason that we namespace the typenames by resource is for related resources -- if you have a child record for instance that is shared across Courses and CourseDetails, but only has a relation on the "courses.v1" resource (possibly using some field that's only available on the Course), we don't want that relationship to show up on the CourseDetails type too.

naptime-graphql/src/test/scala/org/coursera/naptime/ari/graphql/schema/NaptimeUnionFieldTest.scala
bryan-coursera: good catch

PR #186, opened 2017-05-11
Add support for non namespaced typedDefinitions

vkuo-coursera:


naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeUnionField.scala
```diff
@@ -48,8 +49,13 @@ object NaptimeUnionField {
         case definitions: java.util.Map[String @unchecked, String @unchecked] => definitions.asScala
       }.getOrElse(Map[String, String]())
 
-      val unionMemberKey = typedDefinitions
-        .getOrElse(subType.getUnionMemberKey, subType.getUnionMemberKey)
+      val unionMemberKey = (subType, typedDefinitions) match {
```
bryan-coursera: it doesn't look like we actually need to match against `typedDefinitions`, since we're not really doing anything with it that we can't do oyn `typedDefinitions` itself.
I'd probably recommend only matching against subType to handle the different types for that, and always look up in typedDefinition from the different options. We could also do `getOrElse`s to kind of keep trying different things until we find a match in the typedDefinition map, or fall back to the unionMemberKey.

Let's chat more in person maybe

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeUnionField.scala
```diff
@@ -48,8 +49,13 @@ object NaptimeUnionField {
...
+        case (x,y) if y.contains(x.getUnionMemberKey) =>
```
bryan-coursera: style nit: space after `,` and pick better names than `x` and `y`

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeUnionField.scala
```diff
@@ -48,8 +49,13 @@ object NaptimeUnionField {
...
+          typedDefinitions(x.getUnionMemberKey)
+        case (x: RecordDataSchema, y) if y.contains(x.asInstanceOf[RecordDataSchema].getName) =>
+          y(x.asInstanceOf[RecordDataSchema].getName)
```
bryan-coursera: if you're matching against `RecordDataSchema`, you probably don't need to use `.asInstanceOf`

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeUnionField.scala
```diff
@@ -48,8 +50,13 @@ object NaptimeUnionField {
         case definitions: java.util.Map[String @unchecked, String @unchecked] => definitions.asScala
       }.getOrElse(Map[String, String]())
 
-      val unionMemberKey = typedDefinitions
-        .getOrElse(subType.getUnionMemberKey, subType.getUnionMemberKey)
+      val unionMemberKey = subType match {
+        case sType if typedDefinitions.contains(sType.getUnionMemberKey) =>
```
bryan-coursera: we actually don't need to assign `subType` back to anything, since we're not changing its type or anything. so we can actually do:
```scala
case _ if typedDefinitions.contains(subType.getUnionMemberKey) =>
  typedDefinitions(subType.getUnionMemberKey
```

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeUnionField.scala
```diff
@@ -48,8 +50,13 @@ object NaptimeUnionField {
...
+          typedDefinitions(sType.getUnionMemberKey)
+        case sType: NamedDataSchema if typedDefinitions.contains(sType.getName) =>
```
bryan-coursera: and then here, we can call it `namedType` instead of `sType`

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeUnionField.scala
```diff
@@ -48,8 +50,13 @@ object NaptimeUnionField {
...
+          typedDefinitions(sType.getName)
+        case sType => sType.getUnionMemberKey
```
bryan-coursera: and here would be
```scala
case _ => subType.getUnionMemberKey
```

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeUnionField.scala
bryan-coursera: i also think this is only supposed to work if the union's namespace is the same as the child's namespace. can we check against that?

naptime-graphql/src/test/scala/org/coursera/naptime/ari/graphql/schema/NaptimeUnionFieldTest.scala
```diff
@@ -108,4 +108,38 @@ class NaptimeUnionFieldTest extends AssertionsForJUnit with MockitoSugar {
     assert(field.fieldType.toString === expectedField.toString)
   }
 
+  @Test
+  def build_ShorthandTypedDefinitionUnion() = {
```
bryan-coursera: sorry to be obnoxious, but can you modify this test / add one more test that ensures that shorthand syntax doesn't work if the union member is in a different namespace than the parent? (i left a comment above about also modifying the code to support this -- i didn't catch it in the first run, sorry!)

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeUnionField.scala
```diff
@@ -48,8 +50,15 @@ object NaptimeUnionField {
         case definitions: java.util.Map[String @unchecked, String @unchecked] => definitions.asScala
       }.getOrElse(Map[String, String]())
 
-      val unionMemberKey = typedDefinitions
-        .getOrElse(subType.getUnionMemberKey, subType.getUnionMemberKey)
+      val unionMemberKey = subType match {
+        case _ if typedDefinitions.contains(subType.getUnionMemberKey) =>
+          typedDefinitions(subType.getUnionMemberKey)
+        case namedType: NamedDataSchema
```
bryan-coursera: this isn't ever actually going to be a NamedDataSchema though, is it?

vkuo-coursera: I think that it can be, it's the `unionMetadataSchema` that cannot be

bryan-coursera: oh right. my bad! ignore me

naptime-graphql/src/test/scala/org/coursera/naptime/ari/graphql/schema/NaptimeUnionFieldTest.scala
```diff
@@ -42,17 +42,22 @@ class NaptimeUnionFieldTest extends AssertionsForJUnit with MockitoSugar {
...
+      types: List[DataSchema],
+      typedDefinitions: Map[String, String] = Map.empty,
+      properties: Map[String, AnyRef] = Map.empty): UnionDataSchema = {
     val union = new UnionDataSchema()
     val stringBuilder = new java.lang.StringBuilder()
     union.setTypes(types.asJava, stringBuilder)
-    typedDefinitions.foreach(td => union.setProperties(
-      Map("typedDefinition" -> td.asJava.asInstanceOf[AnyRef]).asJava))
+    val modifiableMap = collection.mutable.Map[String, AnyRef]()
+    modifiableMap.getOrElseUpdate("typedDefinition", typedDefinitions.asJava.asInstanceOf[AnyRef])
```
bryan-coursera: why is this `getOrElseUpdate` -- it looks like this is always going to be on an empty map. 

could you instead do:
```scala
val unionProperties = Map("typedDefinition" -> typedDefinitions.asJava.asInstanceOf[AnyRef]) ++ properties
union.setProperties(unionProperties)

File: naptime-graphql/src/test/scala/org/coursera/naptime/ari/graphql/schema/NaptimeUnionFieldTest.scala [hunk 10]
```diff
@@ -108,4 +113,75 @@ class NaptimeUnionFieldTest extends AssertionsForJUnit with MockitoSugar {
...
+          resourceName))))
+    val expectedField = UnionType("courses_v1_typedDefinitionTestField", None, expectedUnionTypes)
+    assert(field.fieldType.toString === expectedField.toString)
+  }
+
+  @Test
+  def build_ShorthandTypedDefinitionUnion_InDifferentNamespace() = {
+    val integerField = new Field(new IntegerDataSchema())
+    integerField.setName("integerField", new java.lang.StringBuilder())
+    val simpleFieldDataSchema = buildRecordField("simpleField", List(integerField), "org.coursera.awaketime")
```
bryan-coursera: that's terrible...

PR #187, opened 2017-05-16
Replace double underscores

vkuo-coursera:


naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/FieldBuilder.scala
```diff
@@ -225,7 +225,7 @@ object FieldBuilder extends StrictLogging {
     * @return GraphQL-safe field name
     */
   def formatName(name: String): String = {
-    name.replaceAll("\\.", "_").replaceAll("/", "_")
+    name.replaceAll("\\.", "_").replaceAll("/", "_").replaceAll("__", "_")
```
bryan-coursera: actually, can we only replace the double underscore if it's at the beginning of the name?

PR #189, opened 2017-05-19
Respect alias names for related resources

bryan-coursera:
When we fetch data for related resources, we must insert it in the datamap based on the alias key, not the field name. This allows us to fetch multiple related resources with different parameters, without overwriting the data based on the original field name

PTAL @yifan-coursera @vkuo-coursera


vkuo-coursera:
LGTM :+1:

PR #190, opened 2017-05-22
Filter out fields without multiget resolvers

vishalkuo:


naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/GraphqlSchemaProvider.scala
```diff
@@ -73,7 +74,10 @@ class DefaultGraphqlSchemaProvider @Inject() (schemaProvider: SchemaProvider)
       }
     }.toMap
     try {
-      val builder = new SangriaGraphQlSchemaBuilder(latestSchema.resources, types)
+      val resources = latestSchema.resources.filter { resource =>
```
vishalkuo: @bryan-coursera I'm having second thoughts about filtering the schema in file vs maybe the `GraphQLController`. It seems sort of gimmicky to me to have an option "filterOnlyMultigets" which is why I placed inside the schema provider, but if you think otherwise, let me know.

bryan-coursera: i agree. i would recommend putting it inside https://github.com/coursera/naptime/blob/master/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/SangriaGraphQlSchemaBuilder.scala#L97 for top-level resources.

and if there's a related resource that uses a forward relation, that should also not be added if there's not a multiget available.

in both of these cases, we should log the error to be able to report these problems better / handle them at a later time

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/GraphqlSchemaProvider.scala
```diff
@@ -23,6 +23,7 @@ import com.linkedin.data.schema.RecordDataSchema
 import com.typesafe.scalalogging.StrictLogging
 import org.coursera.naptime.ari.FullSchema
 import org.coursera.naptime.ari.SchemaProvider
+import org.coursera.naptime.schema.HandlerKind
```
vkuo-coursera: Will remove

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/SangriaGraphQlSchemaBuilder.scala
```diff
@@ -66,7 +66,13 @@ class SangriaGraphQlSchemaBuilder(
       resourceObject <- (try {
         val resourceName = ResourceName(
           resource.name, resource.version.getOrElse(0L).toInt).identifier
-        generateLookupTypeForResource(resourceName)
+        if (!resource.handlers.exists(_.kind == HandlerKind.GET) ||
```
bryan-coursera: i'd add a comment here explaining this logic (it took me a few times to understand what this was checking, and i knew what this logic was supposed to be 😄 )

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeResourceField.scala
```diff
@@ -60,13 +61,34 @@ object NaptimeResourceField {
...
-            resourceName = formatResourceName(resource))
-        }.toList).getOrElse(List.empty)
+          generateField(field, schemaMetadata, resource, schema)
+        }.toList).getOrElse(List.empty).flatten
       }))
     resourceObjectType
   }
 
+  private[schema] def generateField(
+    field: RecordDataSchema.Field,
```
bryan-coursera: these parameters should be four-space indented

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/SangriaGraphQlSchemaBuilder.scala
```diff
@@ -230,8 +233,6 @@ object SangriaGraphQlSchemaBuilder extends StrictLogging {
   }
 
   def scalaTypeToSangria(typeName: String): InputType[Any] = {
-    import sangria.marshalling.FromInput.seqInput
```
bryan-coursera: do you know if these can actually be removed? intellij might be confused here

PR #191, opened 2017-05-24
Fix multiget fields

vishalkuo:


version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.6.9"
+version in ThisBuild := "0.7.0"
```
bryan-coursera: can we make this 0.6.10?

vishalkuo:
This pull request fixes the unnecessary removal of top level resources without `multiget` fields. Essentially resources with no `multiget` but with `finder` fields should still be supported instead of being removed completely.

PR #192, opened 2017-05-26
Refactor GraphQL schema building to capture errors

bryan-coursera:
**This is a pretty large refactor that does the following:**
- Instead of simply logging errors (or throwing exceptions that can be caught later on), most methods were refactored to return SchemaErrors when applicable. There are two ways that this is done:
  - Some methods return a type wrapped with `WithSchemaErrors[T]`, which converts that type into a case class that also includes a list of SchemaErrors generated during the invocation of that call. There are many instances where we want to return some value _and_ a list of errors generated along the way.
  - For cases where its either Error or Value, these have been transformed to use `Either`s instead. Most of these expressions could be represented by an `Either[SchemaError, T]` rather than an `Option[T]`.
- Change resourceNames to be represented by the `ResourceName` class more consistently. This was a requirement for SchemaErrors, since they take in a `ResourceName`, and is generally better than passing Strings everywhere
- Pull a lot of logic around generating TopLevelResourceFields out from the SchemaBuilder, and move it into a new TopLevelResourceField object.
- Some other work to avoid throwing exceptions whenever possible

PTAL @yifan-coursera and @vishalkuo

### DIFF
--- a/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/GraphqlSchemaProvider.scala
+++ b/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/GraphqlSchemaProvider.scala
@@ -20,8 +20,11 @@ import javax.inject.Inject
 
 import com.linkedin.data.schema.RecordDataSchema
 import com.typesafe.scalalogging.StrictLogging
+import org.coursera.naptime.ResourceName
 import org.coursera.naptime.ari.FullSchema
 import org.coursera.naptime.ari.SchemaProvider
+import org.coursera.naptime.ari.graphql.schema.MissingMergedType
+import org.coursera.naptime.ari.graphql.schema.SchemaErrors
 import sangria.schema.Field
 import sangria.schema.ObjectType
 import sangria.schema.Schema
@@ -34,6 +37,7 @@ import scala.util.control.NonFatal
  */
 trait GraphqlSchemaProvider {
   def schema: Schema[org.coursera.naptime.ari.graphql.SangriaGraphQlContext, Any]
+  def errors: SchemaErrors
 }
 
 /**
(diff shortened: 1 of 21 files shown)


naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/GraphqlSchemaProvider.scala
```diff
@@ -97,6 +110,10 @@ class DefaultGraphqlSchemaProvider @Inject() (schemaProvider: SchemaProvider)
     checkSchema()
     cachedSchema
   }
+
+  override def errors: SchemaErrors = {
+    schemaErrors
```
vkuo-coursera: I'm not aware of the standard around this but I've often seen these 'getter' like functions written as 

```
override def errors: SchemaErrors = schemaErrors
```

not sure what the practice around this is.

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/SangriaGraphQlSchemaBuilder.scala
```diff
@@ -56,219 +39,39 @@ class SangriaGraphQlSchemaBuilder(
...
-      Field.apply[SangriaGraphQlContext, DataMap, DataMap, Any](
-        formatResourceTopLevelName(resource),
-        resourceObject,
-        resolve = (context: Context[SangriaGraphQlContext, DataMap]) => {
-          Value(new DataMap())
-        })
+  def generateSchema(): WithSchemaErrors[Schema[SangriaGraphQlContext, DataMap]] = {
+    val topLevelResourceObjectsAndErrors = resources.map { resource =>
+      val lookupTypeAndErrors =
+        NaptimeTopLevelResourceField.generateLookupTypeForResource(resource, schemaMetadata)
```
vkuo-coursera: For my own learning, I'd like to know why we favor static objects over injected dependencies. From my own experience, I've found that it's a lot easier to test when dependencies can be mocked out and static objects should only be used when absolutely necessary. 

i.e. since I have no idea how deep the call graph goes on these static objects, I'd like to use `when` matchers wherever possible.

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeResourceField.scala
```diff
@@ -25,72 +41,84 @@ object NaptimeResourceField extends StrictLogging {
...
+        resourceName <- ResourceName.parse(forwardRelationOption.get)
+        relatedResource <- schemaMetadata.getResourceOpt(resourceName)
+      } yield {
+        if (relatedResource.handlers.exists(_.kind == HandlerKind.MULTI_GET)) {
+          Some(FieldBuilder.buildField(schemaMetadata, field, Option(schema.getNamespace),
+            resourceName = ResourceName.fromResource(resource)))
+        } else {
+          logger.warn(s"Unable to build field for forward relation from " +
+            s"${resource.name} -> ${forwardRelationOption.get} due to the lack of a MULTI_GET handler")
+          None
```
vkuo-coursera: Is this scenario also considered an error? Following along the `HasForwardRelationButNoMultiget` thread

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeTopLevelResourceField.scala
```diff
@@ -0,0 +1,167 @@
...
+      schemaMetadata: SchemaMetadata):
+    WithSchemaErrors[Option[ObjectType[SangriaGraphQlContext, DataMap]]] = {
+
+    val resourceName = ResourceName.fromResource(resource)
+
+    val fieldsAndErrors = resource.handlers
+      .filterNot(handler => MUTATION_HANDLERS.contains(handler.kind))
+      .map { handler =>
+        handler.kind match {
+          // We want to make sure that if a resource has a GET handler, it also has a MULTI_GET
```
vkuo-coursera: Is placing this comment here still relevant? it seems like we've moved this logic down to `generateGetHandler`

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/SangriaGraphQlSchemaBuilder.scala
bryan-coursera: I think you're asking why I'm choosing to making the `generateLookupTypeForResource` static (and put it on an object instead of class). I think there are a few reasons / benefits for doing it this way:
- Separation of concerns: The field generator shouldn't have to maintain its own state of the world. If each field-builder injected in its own schemaMetadata, there's a lot more "state" in the application, and it would be much easier for something to get out of sync or for some oversight while coding. Right now, the SchemaProvider is a singleton and only has one copy of state. If it wasn't a singleton, or had some sort of side effect when loading the schema, or wasn't fully atomic, then different field-builders could technically have different schemas.
- Functional programming: I actually find it much easier to test an individual static function that takes in all of the state it needs, rather than having state part of a class that needs to be initialized. Mocking should work the same way for both classes and functions, but there's less state to worry about. (And no risk of side effects if you call the function multiple times)
- Static objects can be used from other static objects / functions. If the parent calling this was an object, and this was a class, we'd have to convert the parent into a class and inject this in, which could add further complexity.

Let me know if any of this doesn't make sense -- happy to chat more in person :)

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeResourceField.scala
bryan-coursera: yes, good catch!

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeTopLevelResourceField.scala
bryan-coursera: indeed

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/GraphqlSchemaProvider.scala
bryan-coursera: Hmm, I don't know if we have much of a style preference for one over the other.

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/SangriaGraphQlSchemaBuilder.scala
vkuo-coursera: Makes sense, thanks!

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeResourceField.scala
bryan-coursera: i'm actually going to be deprecating this entire section (forward relations aren't used anymore), so instead of refactoring things to get this to work, i'm just going to add a todo to delete the whole section :)

PR #197, opened 2017-06-09
Adds sorting to responses

vishalkuo:


naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimePaginatedResourceField.scala
```diff
@@ -173,9 +173,8 @@ object NaptimePaginatedResourceField {
             idsWithStart.take(limit)
           }.getOrElse(List.empty)
         }
-        objects.collect {
-          case (id, element) if ids.contains(id) => element
-        }.toSeq
+        val objectMap = objects.map { case (id, element) => id -> element }.toMap
+        ids.map(id => objectMap(id))
```
bryan-coursera: i believe this could throw an exception if we somehow have an id that we don't have an object for. I'd recommend making it safer by doing:
```scala
ids.flatMap(id => objectMap.get(id))
```

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimePaginatedResourceField.scala
```diff
@@ -173,9 +173,8 @@ object NaptimePaginatedResourceField {
             idsWithStart.take(limit)
           }.getOrElse(List.empty)
         }
-        objects.collect {
-          case (id, element) if ids.contains(id) => element
-        }.toSeq
+        val objectMap = objects.map { case (id, element) => id -> element }.toMap
```
bryan-coursera: are you able to do `objects.toMap` directly? i thought in scala that the -> was the same as just a 2-tuple

naptime/src/main/scala/org/coursera/naptime/ari/engine/EngineImpl.scala
```diff
@@ -282,8 +282,10 @@ class EngineImpl @Inject() (
 
               // MultiGets return a list of ids, and Gets return a single id (or null)
               val intersection = responseIds.filter(id => ids.contains(id.toString))
+              val indexedIds = ids.zipWithIndex.toMap
+              val sortedIntersection = intersection.sortBy(id => indexedIds(id.toString))
```
bryan-coursera: same comment here -- is it ever possible that id wouldn't be found in indexedIds? might be safer to do `.get` and then a `.getOrElse(-1)` or something

PR #200, opened 2017-07-07
Add resource service to object types

vishalkuo:


naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeTopLevelResourceField.scala
```diff
@@ -68,10 +68,16 @@ object NaptimeTopLevelResourceField extends StrictLogging {
     val fields = fieldsAndErrors.flatMap(_.right.toOption)
     val errors = SchemaErrors(fieldsAndErrors.flatMap(_.left.toOption))
 
+    val resourceService = resource.attributes
+      .find(attribute => attribute.name == "doc")
+      .map(attribute => attribute.value.get.data().getString("service"))
```
bryan-coursera: My initial thought is to make this something more generic, to be able to support any arbitrary attributes (without having to republish naptime or anything). I'd recommend taking an approach where you iterate through the datamap and just print out the keys and values found, perhaps something like:

```scala
val description = resource.attributes
  .find(attribute => attribute.name == "doc")
  .map { attribute =>
    val data = attribute.value.map(_.data()).getOrElse(new DataMap())
    data.keySet().asScala.map { key =>
      val valueStr = Option(data.get(key)).map(_.toString).getOrElse("???")
      s"$key -> $valueStr"
    }
  }
  .mkString("Attributes:\n", "\n", "")
```

(Note: I didn't test that code at all. it's an example)

Another quick note: `value` is an Option, so we should avoid calling `.get` on it. If it's None, that results in an exception :/

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeTopLevelResourceField.scala
```diff
@@ -68,10 +70,22 @@ object NaptimeTopLevelResourceField extends StrictLogging {
...
+    val description = resource.attributes
+      .find(_.name == "doc")
+      .map { attribute =>
+        val data = attribute.data()
+        data.keySet.asScala.map { key =>
+          val valueStr = Option(data.get(key)).map(_.toString).getOrElse("???")
+          s"$key -> $valueStr"
+        }
+      }
+    .mkString("Attributes:\n", "\n", "")
```
bryan-coursera: oh oops, this .mkString should probably be on line 80

naptime-graphql/src/test/scala/org/coursera/naptime/ari/graphql/schema/NaptimeTopLevelResourceFieldTest.scala
```diff
@@ -64,7 +64,10 @@ class NaptimeTopLevelResourceFieldTest extends AssertionsForJUnit {
       "partner",
       "coursePlatform",
       "arbitraryData")
+    val expectedServiceName = Some("Attributes:\nSet(value -> {service=myService}, name -> doc)")
```
bryan-coursera: that doesn't look right 😉 . see comment above

PR #208, opened 2017-08-02
Log parse failures

eeasley-coursera:


bryan-coursera:
@eeasley-coursera can you explain what this is doing? From a quick look at the code, it appears that the entire body of a request is logged on parse failures. Are there security implications of this approach (i.e. logging sensitive cookies or post bodies)? And should something like this be sampled?

eeasley-coursera:
I don't expect it to log the whole body. I just expect it to log this: https://github.com/coursera/naptime/blob/master/naptime/src/main/scala/org/coursera/naptime/actions/RestActionBuilder.scala#L121. When I follow the code, that's what it looks like happens to me. 

This task initially comes from https://coursera.atlassian.net/browse/BACK-1141 which ultimately comes from the fact that I was able to deploy a change that broke prod in an obvious way (JSON deserialization failing) without any automatic systems (datadog, sumologic, etc) surfacing anything.

eeasley-coursera:
Bump

pseudonom:
Bumpump.

PR #213, opened 2017-10-12
Upgrade to Play 2.6

### LINKED ISSUE #73: Always use provided executor context
There are a couple places where we use our own execution context. Instead, we should use a provided execution context.

```
naptime/src/main/scala/org/coursera/naptime/actions/RestAction.scala
62:  import play.api.libs.concurrent.Execution.Implicits.defaultContext

naptime/src/main/scala/org/coursera/naptime/actions/RestActionBuilder.scala
98:      import play.api.libs.concurrent.Execution.Implicits._

naptime-testing/src/main/scala/org/coursera/naptime/actions/RestActionTester.scala
44:      import play.api.libs.concurrent.Execution.Implicits.defaultContext
```

yifan-coursera:
- Separate Play version from Play JSON version in dependencies since they are now independent projects.
- Use Content-Type through `HttpEntity`instead of through headers.
- Create `ResourceTestImplicits` which uses an internal actor system to provide an `ExecutionContext` and `Materializer`, and terminates the actor system after each test.
- Bump versions of Scala, SBT, Courier and Courscala dependencies.
- Replace `Iteratee` with `Accumulator`.
- Replace `ValidationError` with `JsonValidationError`.
- Change `errorRoutes` from `Action` to `EssentialAction` to avoid providing an `ExecutionContext`.
- Fix #73 by removing all uses of `scala.concurrent.ExecutionContext.Implicits.global` and play.api.libs.concurrent.Execution.Implicits.defaultContext and using `Materializer` and `ExecutionContext` provided by application.


yifan-coursera:
cc @zhaojunz because your changes are included in here.

PR #214, opened 2017-10-16
Add timing for query parsing to GraphQLController

bryan-coursera:
This will let us see if there are random slowdowns in query parsing. If not, we'll instrument more!

PTAL @yifan-coursera or @mkovacs-coursera

### DIFF
--- a/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/GraphQLController.scala
+++ b/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/GraphQLController.scala
@@ -118,7 +118,10 @@ class GraphQLController @Inject() (
       variables: JsObject,
       operation: Option[String]): Future[OutgoingQuery] = {
     Future {
-      QueryParser.parse(query) match {
+      val parsedQuery = metricsCollector.timeQueryParsing {
+        QueryParser.parse(query)
+      }
+      parsedQuery match {
         case Success(queryAst) =>
           val baseFilter: IncomingQuery => Future[OutgoingQuery] = (incoming: IncomingQuery) => {
             val context = SangriaGraphQlContext(fetcher, requestHeader, ec, incoming.debugMode)
--- a/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/GraphQLController.scala
+++ b/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/GraphQLController.scala
@@ -118,7 +118,7 @@ class GraphQLController @Inject() (
       variables: JsObject,
       operation: Option[String]): Future[OutgoingQuery] = {
     Future {
-      val parsedQuery = metricsCollector.timeQueryParsing {
+      val parsedQuery = metricsCollector.timeQueryParsing(operation.getOrElse("AnonymousQuery")) {
         QueryParser.parse(query)
       }
       parsedQuery match {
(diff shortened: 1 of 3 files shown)


naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/middleware/MetricsCollectionMiddleware.scala
```diff
@@ -39,14 +39,27 @@ class MetricsCollectionMiddleware(
...
 class LoggingMetricsCollector extends GraphQLMetricsCollector with StrictLogging {
   def markFieldError(fieldName: String): Unit = {
     logger.info(s"Error when loading field $fieldName")
   }
+
+  def timeQueryParsing[A](f: => A): A = {
+    val before = System.currentTimeMillis()
+    val res = f
+    val after = System.currentTimeMillis()
+    logger.info(s"Parsed query in ${after - before}ms")
```
keith-coursera: This message would be much more useful if you passed in and logged `operation` here.  Maybe some graphQL queries are slower than others.

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/GraphQLController.scala
```diff
@@ -118,7 +118,10 @@ class GraphQLController @Inject() (
       variables: JsObject,
       operation: Option[String]): Future[OutgoingQuery] = {
     Future {
-      QueryParser.parse(query) match {
+      val parsedQuery = metricsCollector.timeQueryParsing {
+        QueryParser.parse(query)
+      }
```
keith-coursera: This is a start.  Do you think the later assignments to `incomingQuery`, `filterFn` and the call to `filterFn(incomingQuery)` are worth tracking, too?

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/middleware/MetricsCollectionMiddleware.scala
keith-coursera: Another note about this.  This is going to probably cause a lot of logging.  Consider logging at `debug` level, then we turn it on while triaging?  I'm not sure how you guys normally do this.

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/GraphQLController.scala
bryan-coursera: If this doesn't give us any actionable data, we can add some more tracking in. I'm pretty sure those are all very cheap operations, so I don't want to over-collect data. But it's easy enough to add it in later.

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/middleware/MetricsCollectionMiddleware.scala
bryan-coursera: This MetricsCollector isn't designed to be used in production (we override this with a DatadogMetricsCollector in our repo), so this is mostly just used for debugging in the `example` project

bryan-coursera: ah, but I could tag the metric with operation name. good point

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/GraphQLController.scala
keith-coursera: agreed

PR #215, opened 2017-10-21
relax constraint in CourierFormats.bigDecimalToNumber

mkovacs-coursera:


mkovacs-coursera:
Here's how I tested it: https://phabricator.dkandu.me/D61877

bryan-coursera:
looks good. if there's a unit test that you can put inside naptime, that would be best.

PR #218, opened 2017-11-21
Make small fixes on various complition warnings in Naptime

zhaojunz:
Reviewers: yifan, bryan

CC:

Test Plan: unit test

Auditors:

### DIFF
--- a/naptime/src/main/scala/org/coursera/naptime/actions/RestAction.scala
+++ b/naptime/src/main/scala/org/coursera/naptime/actions/RestAction.scala
@@ -39,6 +39,7 @@ import play.api.mvc.Request
 import play.api.mvc.RequestHeader
 import play.api.mvc.RequestTaggingHandler
 import play.api.mvc.Result
+import play.api.mvc.request.RequestAttrKey
 
 import scala.concurrent.Await
 import scala.concurrent.ExecutionContext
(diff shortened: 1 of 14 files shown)


naptime/src/main/scala/org/coursera/naptime/actions/RestAction.scala
```diff
@@ -231,7 +232,7 @@ trait RestAction[RACType, AuthType, BodyType, KeyType, ResourceType, ResponseTyp
    */
   override def tagRequest(request: RequestHeader): RequestHeader = {
     tags.map { tags =>
-      request.copy(tags = tags)
+      request.addAttr(RequestAttrKey.Tags, tags)
```
bryan-coursera: can you explain this change?

zhaojunz: sure, so  the copy function is deprecated in play 2.6

https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/mvc/RequestHeader.scala#L281

I am just copying the internal impl out to avoid the copy deprecation warning

https://github.com/playframework/playframework/blob/master/framework/src/play/src/main/scala/play/api/mvc/RequestHeader.scala#L302

PR #219, opened 2017-11-21
Remove attribute provider

zhaojunz:
It doesn't seem to be used. not sure why we need to keep them here as it is quite confusing to me the purpose of the code.

In our own codebase, only the constant is used. You can search AttributesProvider in github to find it

Reviewers: yifan, bryan

CC:

Test Plan: no test

Auditors:


zhaojunz:
nvm, it is used in the macro :-( fml.

PR #222, opened 2018-01-24
Add SlowLog middleware to Sangria executor

dguo-coursera:
Add visibility to be able to debug slow queries.


zhaojunz:
why can't we directly use the SlowLog middleware without the wrapper? I am not sure I totally see the value of the wrapper yet.

zhaojunz:
Lgtm

PR #223, opened 2018-01-30
Remove unused Play dependency

yifan-coursera:
`naptime-pegasus` should not depend on Scala Play because it is a pure Java project. In particular, this was adding dependencies on Scala 2.11 packages, which prevented it from being used in a project with any other Scala version.


zhaojunz:
lgtm

PR #232, opened 2018-03-27
GraphQL relation id interpolation to allow for ids as fields nested in arrays of records

cliu587:
@zhaojunz @dguo-coursera @yifan-coursera 

This unifies the id interpolation, and makes it "just work". Meaning if we have records that look like
```
{"courseIds": [1,2,3]}
```
and 
```
{"courses": [{"id": 1, "name": "first"},{"id": 2, "name": "second"}]} 
``` 
we can interpolate the first as `courseIds`, and the second as `courses.id`. This is useful to avoid the need to create extra fields solely for GraphQL relationships.

We also make the single and multiple element interpolation consistent: currently, `course~$id` where `$id` is a single id works, but `course~$ids` where `$ids` is an array does not, which feels unintuitive.

Also a lot of tests were added because a lot of the old behaviors were undocumented and untested.


cliu587:
I thought about this for a bit more: if we need more flexibility we should not be trying to invent our own interpolation language and use something like `jq` syntax [e.g https://github.com/eiiches/jackson-jq] to grab the ids from a record. The only issue is fixing up all the typed definition stuff. I think it might be better to convert the datamap into JSON and then do this, but there's potential performance implications, it's more complicated, ...

This change sits somewhere inbetween the status quo [not consistent, not flexible] and the `jq` solution [consistent, flexible, but less intuitive], so it should be a good starting point.

cliu587:
Gentle ping.

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeResourceUtils.scala
```diff
@@ -158,24 +162,27 @@ object NaptimeResourceUtils extends StrictLogging {
...
-            val fieldName = variableName.split("/").last
-            Utilities.getValueAtPath(
-              data.parentModel.value,
-              data.parentModel.schema,
-              variableName.split("/")).map {
-              case dataList: DataList => dataList.asScala.mkString(",")
-              case other: Any => other.toString
-            }.getOrElse("")
-          })
+        val interpolatedIds: List[String] = InterpolationRegex.findAllMatchIn(value).flatMap { regexMatch =>
```
dguo-coursera: nit: is the type hint required here?

naptime/src/main/scala/org/coursera/naptime/ari/engine/Utilities.scala
```diff
@@ -36,51 +36,94 @@ import scala.collection.JavaConverters._
 
 
 object Utilities extends StrictLogging {
-
-  // TODO(bryan): Dedupe this from NaptimeUnionField
```
dguo-coursera: is this todo still relevant? should we file an issue/jira for it if its still an issue?

naptime/src/main/scala/org/coursera/naptime/ari/engine/Utilities.scala
```diff
@@ -36,51 +36,94 @@ import scala.collection.JavaConverters._
...
-    } yield {
-      typeName
-    }).getOrElse(dataElement.getName.toString)
-    val parentNames = Option(dataElement.getParent).map(replaceTypedDefinitionNames)
-    (parentNames.getOrElse(List.empty) :+ thisName).filterNot(_.isEmpty)
+  private val TYPED_DEFINITION_KEY = "typedDefinition"
+  private def getTypedDefinitionMappings(elements: Iterator[DataElement]): Map[String, String] = {
+    val typedDefinitionMappings = elements.flatMap { element =>
+      Option(element.getSchema.getProperties.get(TYPED_DEFINITION_KEY)).collect {
+        case definitions: java.util.Map[String@unchecked, String@unchecked] => definitions.asScala.toList
```
dguo-coursera: i might be missing something but is there something more efficient we can do here? it seems that we're converting the java map to a scala list only to return a map again?

naptime/src/main/scala/org/coursera/naptime/ari/engine/Utilities.scala
```diff
@@ -36,51 +36,94 @@ import scala.collection.JavaConverters._
...
-    val it = Builder.create(element, schema, IterationOrder.PRE_ORDER).dataIterator()
-    Iterator
-      .continually(it.next)
+      path: Seq[String]): List[String] = {
+
+    def getIterator() = {
+      val dataIterator = Builder.create(element, schema, IterationOrder.PRE_ORDER).dataIterator()
+
+      Iterator
+      .continually(dataIterator.next)
```
dguo-coursera: nit: indentation/style; maybe we should run scalafmt on all naptime projects

version.sbt
```diff
@@ -1,2 +1,2 @@
-version in ThisBuild := "0.9.2-alpha0"
+version in ThisBuild := "0.9.2-alpha1"
```
dguo-coursera: version is now out of date

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeResourceUtils.scala
cliu587: I think it increases readability slightly, so will keep.

naptime/src/main/scala/org/coursera/naptime/ari/engine/Utilities.scala
cliu587: It is no longer relevant, the logic is different from `NaptimeUnionField`

naptime/src/main/scala/org/coursera/naptime/ari/engine/Utilities.scala
cliu587: Yep, since we don't use Arcanist, I propose we use https://github.com/lucidsoftware/neo-sbt-scalafmt to autoformat on compile. I'll do that in a separate diff to not pollute this one.

naptime/src/main/scala/org/coursera/naptime/ari/engine/Utilities.scala
cliu587: There is a potential Java map for each element (`element.getSchema.getProperties.get(TYPED_DEFINITION_KEY))`) which we use `.collect` to grab and turn into a list of maps which is then turned into a single map. 

Let me get rid of the `.toList` part, we don't need to convert to a Scala list. I also added an intermediate variable for clarity.

PR #235, opened 2018-03-28
Deduplicate MultiGet Ids

mbarackman-coursera:
This solves a bug where the resolver fails to properly de-duplicate multiGet ids when combining multiGets from multiple child related elements.

Instead of breaking multiGet ids strings into their constituent ids and de-duplicating the individual ids, it was de-duplicating the entire multiget ids strings for all requests.

so for this request:
```
[
  {teamId: 1, memberIds: [1,2,3]},
  {teamId: 2, memberIds: [1,2,3]},
  {teamId: 3, memberIds: [2,3,4]}
]
```

it would have made these requests

```
members.v1?ids=1,2,3
members.v1?ids=1,2,3
members.v1?ids=2,3,4
```

but the resolver attempts to combine multiple multigets into a single multiget, which in this case produces the following (which is incorrect):

```
members.v1?ids=1,2,2,3,3,4
```

To it’s credit, the backend only return individual member elements, like:

```
[{memberId: 1},{memberId: 2}, {memberId: 3}. {memberId: 4}]
```

But the resolver then iterates over the previously combined id string 1,2,2,3,3,4 post-query to populate the child elements resulting in duplicates in the final response body:

```
[
  {
    "teamId": 1,
    "memberIds": [1,2,3],
    "members": [
      {"memberId": 1},
      {"memberId": 2},
      {"memberId": 2},
      {"memberId": 3},
      {"memberId": 3}
    ]
  },
  {
    "teamId": 2,
    "memberIds": [1,2,3],
    "members": [
      {"memberId": 1},
      {"memberId": 2},
      {"memberId": 2},
      {"memberId": 3},
      {"memberId": 3}
    ]
  },
  {
    "teamId": 3,
    "memberIds": [2,3,4],
    "members": [
      {"memberId": 2},
      {"memberId": 2},
      {"memberId": 3},
      {"memberId": 3},
      {"memberId": 4}
    ]
  }
]
```

This diff both de-dupes the ids before the request is made and de-dupes the ids again post-request to ensure that we end up with:  

```
members.v1?ids=1,2,3,4
```

```
[
  {
    "teamId": 1,
    "memberIds": [1,2,3],
    "members": [
      {"memberId": 1},
      {"memberId": 2},
      {"memberId": 3}
    ]
  },
  {
    "teamId": 2,
    "memberIds": [1,2,3],
    "members": [
      {"memberId": 1},
      {"memberId": 2},
      {"memberId": 3}
    ]
  },
  {
    "teamId": 3,
    "memberIds": [2,3,4],
    "members": [
      {"memberId": 2},
      {"memberId": 3},
      {"memberId": 4}
    ]
  }
]
```

Note: It’s likely not necessary to do the de-duping twice, but it mirrored the original implementation and seemed like it wouldn’t hurt.

### DIFF
--- a/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/resolvers/NaptimeResolver.scala
+++ b/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/resolvers/NaptimeResolver.scala
@@ -143,16 +143,8 @@ class NaptimeResolver extends DeferredResolver[SangriaGraphQlContext] with Stric
               }.toMap
               sourceRequests.map { sourceRequest =>
                 // TODO(bryan): Clean this up
-                val ids = sourceRequest.arguments
-                  .find(_._1 == "ids")
-                  .map(_._2)
-                  .map {
-                    case JsArray(idValues) => idValues
-                    case value: JsValue => List(value)
-                  }
-                  .getOrElse(List.empty)
-
-                val elements = ids.flatMap(parsedElementsMap.get).toList
+                val elements = parseAndDeduplicateIds(sourceRequest)
+                    .flatMap(parsedElementsMap.get).toList
                 val url = successfulResponse.url.getOrElse("???")
                 sourceRequest.idx ->
                   Right[NaptimeError, NaptimeResponse](
@@ -164,7 +156,7 @@ class NaptimeResolver extends DeferredResolver[SangriaGraphQlContext] with Stric
                   Left(NaptimeError(error.url.getOrElse("???"), error.code, error.message))
               }.toMap
           }
-      }
+        }
     }.map(_.flatten.toMap)
   }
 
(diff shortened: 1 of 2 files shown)


naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/resolvers/NaptimeResolver.scala
```diff
@@ -185,15 +177,34 @@ class NaptimeResolver extends DeferredResolver[SangriaGraphQlContext] with Stric
...
         // TODO(bryan): Limit multiget requests by number of ids as well, to avoid http limits
         Request(
           header,
           resourceName,
-          nonIdArguments + ("ids" -> JsArray(ids))) -> innerRequests
+          nonIdArguments + ("ids" -> JsArray(parseAndDeduplicateIds(innerRequests)))) -> innerRequests
       }
   }
 
+  private[this] def parseAndDeduplicateIds(innerRequests: Vector[NaptimeRequest]): Vector[JsValue] = {
```
cliu587: nit: just call this `requests`

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/resolvers/NaptimeResolver.scala
```diff
@@ -185,15 +177,34 @@ class NaptimeResolver extends DeferredResolver[SangriaGraphQlContext] with Stric
...
+    val parsedIds = for {
+      request <- innerRequests
+      id <- parseAndDeduplicateIds(request)
+    } yield id
+    parsedIds.distinct
+  }
+
+  private[this] def parseAndDeduplicateIds(request: NaptimeRequest): Seq[JsValue] = {
+    request.arguments
+      .find(_._1 == "ids")
```
cliu587: I know this was like this before, but can we not use `_._1`, `_._2` far away from the definition location? Here `case (key, _)` and next line `.values` would be a lot more meaningful

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/resolvers/NaptimeResolver.scala
```diff
@@ -143,16 +143,8 @@ class NaptimeResolver extends DeferredResolver[SangriaGraphQlContext] with Stric
...
-                  .find(_._1 == "ids")
-                  .map(_._2)
-                  .map {
-                    case JsArray(idValues) => idValues
-                    case value: JsValue => List(value)
-                  }
-                  .getOrElse(List.empty)
-
-                val elements = ids.flatMap(parsedElementsMap.get).toList
+                val elements = parseAndDeduplicateIds(sourceRequest)
```
cliu587: can we avoid doing this and rename `mergeMultigetRequests` to `mergeMultigetRequestsAndDedupIds`? That way readers don't have to wonder why there's dedup logic in 2 places.

PR #237, opened 2018-04-04
Use Try to be safe to NPEs, and no-op when id interpolation did not find an id

cliu587:
Sorry, found these issues while testing. I haven't released `0.9.2-alpha3` yet so not bumping up the version. I'll merge these changes in after doing more testing locally.

### DIFF
--- a/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeResourceUtils.scala
+++ b/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeResourceUtils.scala
@@ -162,7 +162,8 @@ object NaptimeResourceUtils extends StrictLogging {
 
     relation.arguments
       .mapValues { value =>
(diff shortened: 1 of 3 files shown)


naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeResourceUtils.scala
```diff
@@ -191,4 +199,45 @@ object NaptimeResourceUtils extends StrictLogging {
       }
       .toSet
     }
+
+  private[schema] def interpolate(
+      argumentValue: String,
+      variableNameToInterpolatedIds: Map[String, List[String]]): List[String] = {
+
+    // Note: For simplicity, we do not support cross product of ids. Instead, we
```
cliu587: FYI, here is the logic we are going with when interpolating multiple variables. I don't think right now anyone interpolates multiple variables and multiple ids per variable, so this impl should not affect any existing behaviors.

cliu587:
I tested with a few complicated looking GraphQL queries and things look ok, so I will publish and test in prod today. If there are still issues I'll continue to fix on this diff.

PR #246, opened 2018-04-27
Add scalafmt plugin and apply to project

cliu587:
FYI @dguo-coursera, @yifan-coursera

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3,3 +3,5 @@ scalaVersion in ThisBuild := "2.12.4"
 crossScalaVersions in ThisBuild := Seq("2.11.11", "2.12.4")
 
 organization in ThisBuild := "org.coursera.naptime"
+
+scalafmtOnCompile in ThisBuild := true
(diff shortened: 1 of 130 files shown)


project/plugins.sbt
```diff
@@ -15,3 +15,6 @@ libraryDependencies += { "org.scala-sbt" % "scripted-plugin" % sbtVersion.value
 
 // addSbtPlugin("de.heikoseeberger" % "sbt-header" % "1.5.1")
 addSbtPlugin("de.heikoseeberger" % "sbt-header" % "1.5.1-2-g8b57b53")
+
+addSbtPlugin("com.lucidchart" % "sbt-scalafmt-coursier" % "1.15")
```
dguo-coursera: should we be using `sbt-scalafmt` instead? i wasnt aware able to find coursier being used in this project.

build.sbt
```diff
@@ -3,3 +3,5 @@ scalaVersion in ThisBuild := "2.12.4"
 crossScalaVersions in ThisBuild := Seq("2.11.11", "2.12.4")
 
 organization in ThisBuild := "org.coursera.naptime"
+
+scalafmtOnCompile in ThisBuild := true
```
dguo-coursera: even if we choose to use default formatting, i think itd also be a good idea to keep a copy of scalafmt.conf so we can at a glance see what is being applied (and also able to keep the settings independent of version bumps)

project/plugins.sbt
cliu587: According to https://github.com/lucidsoftware/neo-sbt-scalafmt#usage `if you use coursier, you must use sbt-scalafmt-coursier`

build.sbt
cliu587: Good idea. Let me do that and point to `scalafmt.conf`. We can copy the one from `infra-services` to start.

PR #247, opened 2018-04-27
Fix up macros so that custom request and response types are visible in schemas

amory-coursera:
We're exploring code generation based on Naptime resource schema definitions. While attempting to write generators for clients, I found that some fields in the `Handler` schema were not being generated. This revision fixes TODOs in macro generators, adds an additional `authType` fields (which is important so that clients no about header preconditions), and unwraps / removes the un-used `CustomBodyType.courier` union, which seems to anticipate future features (e.g. arbitrary blob return types) that are unlikely to be built.

cc @mabedin-coursera

### DIFF
--- a/naptime-testing/src/test/scala/org/coursera/naptime/AuthMacroTest.scala
+++ b/naptime-testing/src/test/scala/org/coursera/naptime/AuthMacroTest.scala
@@ -1,5 +1,5 @@
 /*
- * Copyright 2016 Coursera Inc.
+ * Copyright 2018 Coursera Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -36,8 +36,9 @@ object CustomAuthorizer extends HeaderAccessControl[CustomAuth] {
   override private[naptime] def check(authInfo: CustomAuth) = ???
 }
 
-class AuthorizedResource(implicit val executionContext: ExecutionContext,
-                         val materializer: Materializer)
+class AuthorizedResource(
+    implicit val executionContext: ExecutionContext,
+    val materializer: Materializer)
     extends TopLevelCollectionResource[String, Item] {
 
   override def keyFormat: KeyFormat[String] = KeyFormat.stringKeyFormat
@@ -49,7 +50,8 @@ class AuthorizedResource(implicit val executionContext: ExecutionContext,
   implicit val fields = Fields.withDefaultFields("name")
 
   def get(id: String) =
-    Nap.auth(CustomAuthorizer)
+    Nap
+      .auth(CustomAuthorizer)
       .get { ctx =>
         ???
       }
@@ -60,17 +62,14 @@ object AuthorizedResource {
   val routerBuilder = Router.build[AuthorizedResource]
 }
 
-class AuthMacroTest
-    extends AssertionsForJUnit
-    with MockitoSugar
-    with ResourceTestImplicits {
+class AuthMacroTest extends AssertionsForJUnit with MockitoSugar with ResourceTestImplicits {
 
   val schema = AuthorizedResource.routerBuilder.schema
 
   @Test
   def get(): Unit = {
     val handler = schema.handlers.find(_.name === "get").get
-    assert(handler.authType.contains("org.coursera.naptime.CustomAuth"))
+    assert(handler.authType === Some("org.coursera.naptime.CustomAuth"))
   }
 
 }
--- a/naptime-testing/src/test/scala/org/coursera/naptime/AuthMacroTest.scala
+++ b/naptime-testing/src/test/scala/org/coursera/naptime/AuthMacroTest.scala
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2016 Coursera Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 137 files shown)


naptime-testing/src/test/scala/org/coursera/naptime/AuthMacroTest.scala
```diff
@@ -0,0 +1,76 @@
+/*
+ * Copyright 2016 Coursera Inc.
```
mabedin-coursera: 2018?

amory-coursera: Good point - I had copy / pasted from a previous file. Will fix.

naptime/src/main/scala/org/coursera/naptime/router2/MacroImpls.scala
```diff
@@ -146,9 +146,9 @@ class MacroImpls(val c: blackbox.Context) {
         case (tpe, methods) if ACTION_PATCH =:= tpe =>
           buildPatchTree(methods)
         case (tpe, methods) if ACTION_FINDER =:= tpe =>
-          buildFinderTree(methods, tpe)
```
cliu587: thank you for removing the unused variables!

naptime/src/main/scala/org/coursera/naptime/router2/MacroImpls.scala
```diff
@@ -460,14 +461,42 @@ class MacroImpls(val c: blackbox.Context) {
             org.coursera.courier.templates.DataTemplates.DataConversion.SetReadOnly)
         """
       }
-      // TODO: handle input, custom output bodies, and attributes
+
+      val vagueTypes = List("Unit", "play.api.mvc.AnyContent", "Any")
```
cliu587: do you mind adding a comment what `vague` means here? is it just that we'd rather not return `Some[...]` for the input body/custom output bodies for these types, which are 'defaults' if not user specified?

naptime-testing/src/test/scala/org/coursera/naptime/AuthMacroTest.scala
```diff
@@ -0,0 +1,76 @@
...
+    extends AssertionsForJUnit
+    with MockitoSugar
+    with ResourceTestImplicits {
+
+  val schema = AuthorizedResource.routerBuilder.schema
+
+  @Test
+  def get(): Unit = {
+    val handler = schema.handlers.find(_.name === "get").get
+    assert(handler.authType.contains("org.coursera.naptime.CustomAuth"))
```
cliu587: nit/style: maybe `authType = Some("org.coursera.naptime.CustomAuth")` is a slight improvement? subjective and feel free to ignore.

naptime-testing/src/test/scala/org/coursera/naptime/NonNestedMacroTests.scala
```diff
@@ -169,6 +169,13 @@ class NonNestedMacroTests extends AssertionsForJUnit with MockitoSugar with Reso
     assert(taggedRequest.tags.get(Router.NAPTIME_METHOD_NAME).contains(methodName))
   }
 
+  private[this] def checkNoCustomTypes(methodName: String): Unit = {
```
cliu587: nit: let's name this `noCustomInputOutputAuthTypes`?

naptime-testing/src/test/scala/org/coursera/naptime/NonNestedMacroTests.scala
```diff
@@ -246,16 +253,28 @@ class NonNestedMacroTests extends AssertionsForJUnit with MockitoSugar with Reso
...
-    verify(instance).parameterless()
+  def noCustomTypes(): Unit = {
+    checkNoCustomTypes("getAll")
+    checkNoCustomTypes("get")
+    checkNoCustomTypes("multiGet")
+    checkNoCustomTypes("create")
+    checkNoCustomTypes("update")
+    checkNoCustomTypes("delete")
+    checkNoCustomTypes("complex")
+    checkNoCustomTypes("actionWithoutParams")
```
cliu587: what about `actionWithParams`? 

Also, I think the reason some of these handlers have no custom types is because of `???` and others because they are not supposed to have custom types, maybe make that distinction clear? Or I could be missing the code completely, let me know!

naptime/src/main/scala/org/coursera/naptime/router2/MacroImpls.scala
amory-coursera: 👍 thanks for the request, added.

naptime-testing/src/test/scala/org/coursera/naptime/AuthMacroTest.scala
amory-coursera: `IntelliJ` prefers `handler.authType.contains(...` over `handler.authType == Some(...`, but is fine with `handler.authType === Some(...`. I've gone with the third.

naptime-testing/src/test/scala/org/coursera/naptime/NonNestedMacroTests.scala
amory-coursera: Auth type exists only if there is a `.auth` statement in the builder. For `get`, `multiGet` and `finder`, there is no notion of custom request or body type. For `create` and `update`, custom input body types are allowed, but only if explicitly declared with a`.jsonBody` statement. Same for `action`, with the addition that `action[ResponseType]` sets a custom response type.

So, nothing to do with the `???`: it's all about the `Nap` builders are called.

amory-coursera:
Thanks for the review @cliu587 . I'll rebase against @mbarackman-coursera 's changes, bump the version before landing.

PR #248, opened 2018-05-02
Introduce Related Resource Auth Override

mbarackman-coursera:
- This change allows API developers to create related resources that are fetched with “internal” auths. 

- This is done by adding an authOverride value to the reverse relation definition on the primary resource. This will show up in the naptime resource schema and will be interpreted by assembler in order to request with the proper auth request headers.

- The purpose of this is to give developers the option to do authentication in only one place, the entry/parent resource, and use internal auth for all the related/child resources. This allows the child resources to be a bit more re-usable in that they can be attached to multiple parent resources that may all have different auth schemes.

### DIFF
--- a/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/resolvers/NaptimeResolver.scala
+++ b/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/resolvers/NaptimeResolver.scala
@@ -10,6 +10,7 @@ import org.coursera.naptime.ari.graphql.SangriaGraphQlContext
 import org.coursera.naptime.ari.graphql.schema.DataMapWithParent
 import org.coursera.naptime.ari.graphql.schema.NaptimeResourceUtils
 import org.coursera.naptime.ari.graphql.schema.ParentModel
+import org.coursera.naptime.schema.AuthOverride
 import play.api.libs.json.JsArray
 import play.api.libs.json.JsNull
 import play.api.libs.json.JsValue
@@ -25,7 +26,8 @@ case class NaptimeRequest(
     resourceName: ResourceName,
     arguments: Set[(String, JsValue)],
     resourceSchema: RecordDataSchema,
-    paginationOverride: Option[ResponsePagination] = None)
+    paginationOverride: Option[ResponsePagination] = None,
+    authOverride: Option[AuthOverride] = None)
 
 case class NaptimeResponse(
     elements: List[DataMapWithParent],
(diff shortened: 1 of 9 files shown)


naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/resolvers/NaptimeResolver.scala
```diff
@@ -268,4 +296,10 @@ class NaptimeResolver extends DeferredResolver[SangriaGraphQlContext] with Stric
   }
 
 
+  private[this] def getInternalRequestContext(context: SangriaGraphQlContext): SangriaGraphQlContext = {
+    val originalRequestHeader = context.requestHeader
+    val newHeaders = context.requestHeader.headers.remove("X-Coursera-User").add("X-Coursera-Source" -> "internal")
```
cliu587: is it possible to move this to assembler's RemoteFetcher? that way implementation detail of how coursera handles auth isn't leaked onto the Naptime project and I believe that is the place where we have the final headers and instantiate the client so it feels like a more natural place to do this conditional logic.

mbarackman-coursera: yeah I think that's doable and that all sounds reasonable. let me see

naptime/src/main/pegasus/org/coursera/naptime/schema/ReverseRelationAnnotation.courier
```diff
@@ -20,4 +20,12 @@ record ReverseRelationAnnotation {
    */
   relationType: enum RelationType { FINDER, MULTI_GET, GET, SINGLE_ELEMENT_FINDER }
 
+  /**
+   * This is an optional field which will override the auth headers from the
```
cliu587: let's fix up this comment to not refer to headers. indicating that a different auth strategy from the default behavior I think is fine.

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/resolvers/NaptimeResolver.scala
```diff
@@ -86,19 +98,26 @@ class NaptimeResolver extends DeferredResolver[SangriaGraphQlContext] with Stric
...
       .map {
         case (resourceName, requests) =>
+          // Handle MultiGet and Non-Multigets differently, since multigets can be batched
           val (forwardRequests, reverseRequests) =
             requests.partition(_.arguments.exists(_._1 == "ids"))
 
-          // Handle MultiGet and Non-Multigets differently, since multigets can be batched
-          val forwardRelations =
-            fetchForwardRelations(forwardRequests, resourceName, ctx)
+          // partition forward requests by auth type
```
cliu587: nit: the comment should reflect that we need to make 1 call of `fetchForwardRelations` for each auth override type.

PR #249, opened 2018-05-03
Fix bad rebase

cliu587:
Remove `throw`


mbarackman-coursera:
It looks like you're missing the change from your prior commit with the "passthrough" comment. https://github.com/coursera/naptime/commit/fae3a22a41324d8dd0fe21512ad9b3b34b5c5c78#diff-26ff45e2336e9abb37e5dd6f22256ffdR232

I'd recommend creating a new branch off your previous commit (62379bb2ad36e51a7a0f42b7b27080a166a280f8), re-add your scalafmt stuff, run it, then cherry-pick my commit. That should get us to the point where we should be because I don't really know what branch you rebased your original PR off of.

cliu587:
Good call, will try a separate approach.

PR #250, opened 2018-05-04
Fix bad rebase, take 2

cliu587:
cc @mbarackman-coursera 

I checked `naptime/src/main/scala/org/coursera/naptime/ari/engine/Utilities.scala` and `naptime/src/main/scala/org/coursera/naptime/errors.scala` and they look fixed. The diff also does not touch `...resolvers/NaptimeResolver.scala` at all which is what your future diff did to change auth, so I think this is the correct behavior.


mbarackman-coursera:
Looks good to me. Thanks for the clean-up.

PR #251, opened 2018-05-14
Add testActionPassAuth to facilitate testing with trivial auth mocks

amory-coursera:
Years ago, it was common practice to write resource-level unit tests for Coursera APIs - there are 555 usages of the existing `testAction` method in our codebase. As authorization logic became more complicated, teams increasingly moved away from resource testing because it was cumbersome to mock or fake auths. 

The irony here is that `testAction` itself requires the caller define a RestContext with authorization data, and tests only execute the authorizer's test-only `check` method. Needing to define a fully functional authorizer for testing is silly, because the bulk of its machinery won't even be exercised within a `testAction` execution. 

This revision adds a new test method that skips the `check` step of `testAction`. This makes it possible to construct test resources with trivial authorizers - e.g. `mock[<AuthorizerType]`. While we don't add very many _new_ resource-level tests, the option to switch to `testActionPassAuth` when refactoring tests for legacy code will make it _much_ easier to migrate authorizer definitions for legacy code.

### DIFF
--- a/naptime/src/main/scala/org/coursera/naptime/router2/MacroImpls.scala
+++ b/naptime/src/main/scala/org/coursera/naptime/router2/MacroImpls.scala
@@ -88,7 +88,8 @@ class MacroImpls(val c: blackbox.Context) {
    * @tparam Resource The resource type that we are specializing.
    * @return A [[c.Tree]] corresponding to a [[ResourceRouterBuilder]].
    */
-  def build[Resource <: CollectionResource[_, _, _]](implicit wtt: WeakTypeTag[Resource]): c.Tree = {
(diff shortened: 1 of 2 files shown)


naptime/src/main/scala/org/coursera/naptime/router2/MacroImpls.scala
```diff
@@ -88,7 +88,8 @@ class MacroImpls(val c: blackbox.Context) {
    * @tparam Resource The resource type that we are specializing.
```
amory-coursera: Changes to this file from pre-compile `scalafmt` application. Please ignore.

PR #252, opened 2018-05-14
Add body-dependent authorization capability to Rest action handler

amory-coursera:
Coursera authorization rules are often context-dependent. For example, only university staff can create and modify courses for a particular university. When this context is present as a rest handler method argument (e.g. `get(id: CourseId) = Rest ...` or `update(id: CourseId) = Rest ...`), it's straightforward to define context-aware authorizers (e.g. `Rest.auth(canManageCourse(id)...`). Unfortunately, however, we frequently put necessary authorization context data into request bodies (e.g. `create() = Rest.jsonBody[Course].create { ... }`), and in these cases authorization logic must be implemented within handler bodies. 

In the Coursera code base there are 207 applications for `runAndCheck` (authorizers used outside of an `auth` block) and 307 instances of `throw Errors.Forbidden` (ad-hoc authorization application). These non-standard usages are problematic because they make it harder to know what a resource's behavior will be from outside, both as a developer and for automated tooling.

This revision modifies `RestAction` to accept either a simple `HeaderAccessControl[AuthType]` authorizer, or a body-dependent `BodyType => HeaderAccessControl[AuthType]` authorization expressions. This should make it possible for us to standardize many of our non-standard auth applications.

Note that, at least for Coursera, this change is not backward compatible, since our private code base has its own implementations of `RestActionBuilder`, which will need to be updated.

### DIFF
--- a/naptime/src/main/scala/org/coursera/naptime/actions/DefinedBodyTypeRestActionBuilder.scala
+++ b/naptime/src/main/scala/org/coursera/naptime/actions/DefinedBodyTypeRestActionBuilder.scala
@@ -38,7 +38,7 @@ class DefinedBodyTypeRestActionBuilder[
     ResourceKeyType,
     ResourceType,
     ResponseType] private[actions] (
-    authGeneratorOrAuth: AuthGeneratorOrAuth[BodyType, AuthType],
+    authGeneratorOrAuth: AuthGenerator[BodyType, AuthType],
     bodyParser: BodyParser[BodyType],
     errorHandler: PartialFunction[Throwable, RestError])(
     implicit keyFormat: KeyFormat[ResourceKeyType],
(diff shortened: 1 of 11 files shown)


naptime/src/main/scala/org/coursera/naptime/actions/RestActionBuilder.scala
```diff
@@ -67,16 +85,42 @@ class RestActionBuilder[RACType, AuthType, BodyType, ResourceKeyType, ResourceTy
     new RestActionBuilder(auth, bodyParser, errorHandler)
 
   /**
-   * Set the body type.
+   * Like [[auth]] above, but with a body-aware generator function.
+   *
+   * Note that this version of [[auth]] converts the builder into a
+   * [[DefinedBodyTypeRestActionBuilder]], which forbids request body definition changes.
+   */
+  def bodyAuth[NewAuthType](authGenerator: BodyType => HeaderAccessControl[NewAuthType])
```
amory-coursera: As suggested in our private repo by @mbarackman-coursera , it's may be best to only allow allow this from `DefinedBodyTypeRestActionBuilder`, as default body is not meaningful. Thoughts from other reviewers?

mkovacs-coursera: That restriction makes sense to me.

naptime/src/main/scala/org/coursera/naptime/actions/DefinedBodyTypeRestActionBuilder.scala
```diff
@@ -0,0 +1,114 @@
...
+    extends RestActionBuilderTerminators[
+      RACType,
+      AuthType,
+      BodyType,
+      ResourceKeyType,
+      ResourceType,
+      ResponseType] {
+
+  /**
+   * Like [[auth]] above, but with a body-aware generator function.
```
mkovacs-coursera: nit: update comment

PR #255, opened 2018-05-23
Materialize model during JSON deserialization

yifan-coursera:
Previously, deserializing JSON to a courier model resulted in a model that would have lazily initialized fields. This means that if one of the fields was a custom Scala class with custom validation logic in the constructor or coercer, that validation logic would be ignored during deserialization time. Instead, it would run the first time the field was accessed, leading to exceptions thrown at unexpected places. This change forces materialization of the entire model during JSON deserialization time, and causes JSON deserialization to fail if an exception is thrown by the Scala constructor or coercer.


amory-coursera:
Looks good to me. This will really help. 

It's likely that we have materialized data instances with invalid field values that slipped through without a check like this. I think we should go ahead with this and enforce validations regardless, but let's make sure to announce this change when it lands in infra-services, and let folks know what they should do if records become unreadable.

PR #262, opened 2018-09-25
Fixing union field builder if member is a typeref.

ptbarthelemy:
PTAL @cliu587


naptime/src/test/pegasus/org/coursera/naptime/ari/graphql/models/UnionTypeWithTyperef.courier
```diff
@@ -0,0 +1,12 @@
+namespace org.coursera.naptime.ari.graphql.models
+
+/**
+ * The goal of this type is to exercise functionality for union type that contains a record and a
```
cliu587: let's note that `CourseId` is the typeref here to make it clear on readers

naptime/src/test/pegasus/org/coursera/naptime/ari/graphql/models/FakeModel.courier
```diff
@@ -0,0 +1,9 @@
+namespace org.coursera.naptime.ari.graphql.models
+
+/**
+ * A fictional model used for testing purposes
+ */
+record FakeModel {
```
cliu587: Looks like this model is only used for the `unionField` capability, to be explicit can we name it something like `ModelWithTypeRefInUnion`?

naptime-graphql/src/test/scala/org/coursera/naptime/ari/graphql/schema/TestingFetcherApi.scala
```diff
@@ -0,0 +1,31 @@
...
+import org.coursera.naptime.ari.Response
+import play.api.libs.json.JsString
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.Future
+
+/**
+ * An implementation of FetcherApi that allows one to store data keyed by resource name and ID.
+ */
+case class TestingFetcherApi(rawData: TestingFetcherApi.TestingFetcherApiData) extends FetcherApi {
```
cliu587: to be precise with terminology, let's call this `FakeFetcherApi`.

(the terminology first was introduced in http://xunitpatterns.com/Test%20Double.html and then popularized in https://martinfowler.com/articles/mocksArentStubs.html, specifically,
`Fake objects actually have working implementations, but usually take some shortcut which makes them not suitable for production (an in memory database is a good example).`)

naptime-graphql/src/test/scala/org/coursera/naptime/ari/graphql/schema/ExecutorHelper.scala
```diff
@@ -0,0 +1,63 @@
...
+import sangria.schema.Schema
+
+import scala.concurrent.Await
+import scala.concurrent.ExecutionContext
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.duration.Duration
+
+class ExecutorHelper {
+
+  def executeQuery(
```
cliu587: this doesn't look super general, let's make the naming reflect that. Maybe `TestExecutorHelper` to start? sorry I don't have a better suggestion here :(

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeUnionField.scala
```diff
@@ -80,9 +82,22 @@ object NaptimeUnionField {
...
           if (unionDataSchema.getProperties.containsKey(TYPED_DEFINITION_KEY)) {
-            Option(context.value.element.getDataMap("definition"))
+            val resultAsDataMap: Option[Action[SangriaGraphQlContext, Any]] = Try(
+              context.value.element
+                .getDataMap("definition")).toOption
               .map(element => context.value.copy(element = element))
-              .getOrElse(subTypeField.resolve(context))
+            resultAsDataMap.getOrElse {
+              // Create a separate field to access the non data-map directly. This field will be
+              // found at the "definition" key.
```
cliu587: let's add the `why` for this comment, I think the previous impl assumed each union member is a data map which may not be true? e.g a union where one member is a string or enum may make sense?

naptime-graphql/src/test/scala/org/coursera/naptime/ari/graphql/schema/TestingFetcherApi.scala
ptbarthelemy: Great comment, thanks!

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/schema/NaptimeUnionField.scala
ptbarthelemy: I've simplified the code and updated the documentation. Let me know if it makes more sense.

ptbarthelemy:
> To make sure I'm not misunderstanding - the main issue is unions with `@typedDefinition` annotations previously assumed each member can be converted to a DataMap?

I wouldn't describe it exactly this way. It looks like we did expect the field to be a non-`DataMap`, which is why the code that is replaced was `subTypeField.resolve`. However, this `resolve` function did not work correctly because it could not find the field in question.

Consider the test case with field `unionField`. `subTypeField` defined this field as existing in some object keyed on `unionField`. So, its resolve method wouldn't work on `{"typeName": "course", "definition": "courseAId"}`. It seems like in the `DataMap` case we were doing the right thing because we were getting the child at the value `definition`.

> If that's the case then things like enums wouldn't have worked previously either, right?

Yeah, I've added a test to demonstrate that it's working now.

PR #265, opened 2018-12-11
add ability to create NaptimeActionExceptions with a cause

fyi-coursera:
Often in errorHandlers, we catch exceptions and throw the appropriate `NaptimeActionException` to return the correct httpCode to the client. There is currently no way to specify a cause of the `NaptimeActionException` we throw which means we lose the original exception's stack trace.

This change allows you to specify a cause when creating a `NaptimeActionException` which will preserve the stack trace of the original exception and help with debugging.

### DIFF
--- a/naptime/src/main/scala/org/coursera/naptime/actions/RestActionBuilder.scala
+++ b/naptime/src/main/scala/org/coursera/naptime/actions/RestActionBuilder.scala
@@ -167,7 +167,8 @@ class RestActionBuilder[RACType, AuthType, BodyType, ResourceKeyType, ResourceTy
               Status.BAD_REQUEST,
               Some("request.validation"),
               Some(e.getMessage),
-              None)
+              None,
+              Some(e))
             Left(resp.result)
           case NonFatal(e) =>
             logger.error(s"Unknown exception while parsing body of request.", e)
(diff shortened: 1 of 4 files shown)


naptime/src/main/scala/org/coursera/naptime/actions/RestActionBuilder.scala
```diff
@@ -167,7 +167,8 @@ class RestActionBuilder[RACType, AuthType, BodyType, ResourceKeyType, ResourceTy
               Status.BAD_REQUEST,
               Some("request.validation"),
               Some(e.getMessage),
-              None)
+              None,
```
cliu587: Let's follow the boy(or girl) scout rule of leaving the code better than you found it and use named parameters, especially for the `None` which can be anything. Also as we add more parameters it becomes a better idea to name at least some of them.

naptime/src/main/scala/org/coursera/naptime/errors.scala
```diff
@@ -35,8 +35,9 @@ case class NaptimeActionException(
     httpCode: Int,
     errorCode: Option[String],
     message: Option[String],
-    details: Option[JsValue] = None)
-    extends RuntimeException(s"Naptime error $httpCode [$errorCode]: $message") {
+    details: Option[JsValue] = None,
+    cause: Option[Throwable] = None)
+    extends RuntimeException(s"Naptime error $httpCode [$errorCode]: $message", cause.orNull) {
```
cliu587: should the responsibility of turning `cause` to a reasonable stack trace live in this case class as well?

PR #270, opened 2019-06-25
Take in middleware factory rather than middleware

cliu587:
PTAL @dguo-coursera, @amory-coursera - I think middlewares should be created per request rather than once on instantiation, as they hold execution state. This along with another change should unbreak the datadog tracing.

### DIFF
--- a/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/GraphQLController.scala
+++ b/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/GraphQLController.scala
@@ -59,7 +59,7 @@ class GraphQLController @Inject()(
     fetcher: FetcherApi,
     filterList: FilterList,
     metricsCollector: GraphQLMetricsCollector,
-    additionalMiddlewares: List[Middleware[Any]])(implicit ec: ExecutionContext)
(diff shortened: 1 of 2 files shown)


cliu587:
not necessary :) thanks for the 2nd pair of eyes on this, @amory-coursera

cliu587:
Reviving this - while middlewares are indeed stateless, reusing the same middleware for every request seemed to have resulted in a memory leak due to the thread-local scope data not getting cleaned up. 

Here's the heap dump for the newest deploy that resulted in java heap shooting up rapidly: https://imgur.com/a/t5SH0CV (~11GB heap)
 
here's the heap dump for the last known good deploy (which still has tracing enabled):
https://imgur.com/a/cWpNnzu (~700MB heap)

My best guess is because we create a singleton tracing middleware, every thread-local created has a strong reference that's never cleaned up at the end of the request.

cliu587:
Gentle ping on this - would like to resolve by end of week if at all possible. Thanks in advance!

cliu587:
> This feels like something we could fix in another way, but if this is the path of least resistance I'm ok with the approach.

I think the existence of thread-locals in middlewares along with async execution makes all other solutions very non-obvious. Assuming this fix indeed does the job, I think it is the path of least resistance and the most obviously correct solution.

PR #271, opened 2019-07-17
Upgrade to Play 2.7

amory-coursera:
This PR upgrades Naptime to use Play 2.7. Changes in this revision address two incompatibilities, which in turn introduce mild breaking changes in Naptime, motivating a version bump to 0.10.0:

- Play 2.7 has fully deprecated request tags, and instead guides developers to use typed request attributes. This revision shifts all tagging to Play's preferred attribute representation.
- From version 2.5, Play has strongly discouraged static references to the running application via the `Play.maybeApplication` accessor, and instead recommends that the application be injected as a dependency. While applications can be configured to allow static application references by setting `play.allowGlobalApplication=true` in configuration, I think the Play team's advice is good here, and have chosen to follow it. Injecting the application reduces the implicit constructor signature of resources and handlers, and ensures consistent execution contexts.

We'll need to take care in rolling this out. Switching to Application injection rather than materializer + execution context in Coursera apps should be pretty straightforward, but an incautious shift from tags to attrs could break things libraries aren't in sync.


dguo-coursera:
This looks good to me.

PR #277, opened 2020-04-16
Quiet noisy logs when deserializing a NULL as we appear to be handling that case explicitly

cliu587:
See added comment. It appears we handle this case explicitly (`jsObjectToRecord` is what is calling `jsValueToData`) so maybe this check is somewhat over-eager. 

Specifically, L544-550 of this file.
```
 if (field != null) {
          val result = jsValueToData(newSchemaPath, value, field.getType)
          if (result != Null.getInstance) {
            Some(field.getName -> result)
          } else {
            None
          }
```

Also, note that the check is only a log and not a throw, if something terrible happens downstream, we'd still get the same error behaviors.

### DIFF
--- a/naptime-models/src/main/scala/org/coursera/naptime/courier/CourierFormats.scala
+++ b/naptime-models/src/main/scala/org/coursera/naptime/courier/CourierFormats.scala
@@ -462,7 +462,8 @@ object CourierFormats extends StrictLogging {
       schema: DataSchema): AnyRef = {
     (schema, jsValue) match {
       case (_, JsNull) =>
-        checkSchema[NullDataSchema](schemaPath, "deserializing a NULL", schema)
+        // We don't check the schema when deserializing a NULL, since we can encode any optional schema into NULL.
+        // See `jsObjectToRecord` for where we set com.linkedin.data.Null to `None`...
         Null.getInstance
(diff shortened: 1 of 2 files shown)


cliu587:
A little more context:

we started seeing a bunch of 
```
2020-03-10 23:44:48,313 WARN [IMmz_WMpEeqnmRJziz0ezw] [6779443058966428034] [o.c.n.c.CourierFormats$] [-] >>> Schema mismatch in NewCourierFormats when deserializing a NULL in org.coursera.playcour.requestcontext.RequestContextUserProperties: expected schema of type com.linkedin.data.schema.NullDataSchema, but got com.linkedin.data.schema.ArrayDataSchema
```

Which was traced back to my diff in https://coursera.slack.com/archives/C0259PQKA/p1586980493062000 - I think this is because assembler's request context is missing some data for the optional fields as compared to edge's, but I don't think this is a cause for concern, considering this has been running in prod for a month or so, and the log doesn't do any actual error raising.

cliu587:
Screen Shot 2020-04-15 at 9 26 12 PM

Yeah, we had this before as well. It was not as frequent. I'll do a bit of a deeper dive to understand what the root cause is, since the diff we suspect now seems to only apply to trace context, not request context (https://github.com/webedx-spark/infra-services/commit/fec757a110de7a45d8dbbffa29d9377ba8259373)

PR #278, opened 2020-07-15
Use Matcher.quoteReplacement to escape ids with special characters

cliu587:
Should fix the issue described in https://coursera.slack.com/archives/C025SQ0GH/p1594832573018600 which comes from ids with `$` embedded. Looks like the root cause is the replacement text for the `regex. replaceAllIn` method needs to escape special characters. They provide an API for this.

Existing unit tests should cover regressions, and added a new unit test for this case.

This is only deployed through assembler so fairly safe to deploy with traffic shifting + E2E tests as well.


sehranhas:
LGTM

cliu587:
Gentle ping, thanks in advance!

PR #280, opened 2020-08-25
Add includes relations to dataSchema

ppaskaris-coursera:
This PR adds Naptime include relations to the dataSchema, similar to how we stuff the GraphQL relations in as pseudo-fields. In particular, we add a new field property `"included"` with is an object with a `"resourceName"` property. When we have both a GraphQL and Naptime relation, both the `"relatedOn"` and `"included"` properties are present, otherwise only the relation that applies for that pseudo-field is present. Consider the resource with fields:

```
ResourceFields
  .withRelated(
    "legacyRelation" -> ResourceName("first", 1),
    "modernRelation" -> ResourceName("second", 2))
  .withGraphQLRelations(
    "modernRelation" -> GetGraphQLRelation(ResourceName("second", 2), "relatedId"))
```

It produces a fields array on the model for this resource like this:

```
{
  "fields": [
    {
      "name": "legacyRelation",
      "included": {
        "resourceName": "first.v1"
      },
      "optional": false,
      "type": {
        "type": "array",
        "items": "string"
      }
    },
    {
      "name": "modernRelation",
      "included": {
        "resourceName": "second.v2"
      },
      "relatedOn": {
        "resourceName": "second.v2",
        "id": "relatedId"
      },
      "optional": true,
      "type": "string"
    }
  ]
}
```

### DIFF
--- a/naptime/src/main/scala/org/coursera/naptime/Types.scala
+++ b/naptime/src/main/scala/org/coursera/naptime/Types.scala
@@ -22,6 +22,7 @@ import com.linkedin.data.schema.ArrayDataSchema
 import com.linkedin.data.schema.DataSchema
 import com.linkedin.data.schema.MapDataSchema
 import com.linkedin.data.schema.Name
+import com.linkedin.data.schema.NullDataSchema
 import com.linkedin.data.schema.PrimitiveDataSchema
 import com.linkedin.data.schema.RecordDataSchema
 import com.linkedin.data.schema.RecordDataSchema.RecordType
@@ -40,7 +41,7 @@ object Types extends StrictLogging {
 
   object Relations {
     val RELATION_PROPERTY_NAME = "relatedOn"
-    val includedPropertyName = "included"
+    val INCLUDED_PROPERTY_NAME = "included"
   }
 
   @deprecated("Please use the one with fields included", "0.2.4")
@@ -92,8 +93,8 @@ object Types extends StrictLogging {
       case unknown: DataSchema =>
         throw new RuntimeException(s"Cannot compute asymmetric type for key type: $unknown")
     }
-    val relationKeys = fields.relations.keys ++ fields.graphQLRelations.keys
-    for (name <- relationKeys) {
+    val relationNameSet = fields.relations.keySet ++ fields.graphQLRelations.keySet
+    for (name <- relationNameSet) {
       val relationOption = fields.relations.get(name)
       val graphQLRelationOption = fields.graphQLRelations.get(name)
       if (mergedSchema.contains(name)) {
@@ -103,9 +104,13 @@ object Types extends StrictLogging {
       } else {
         val nameSegments = name.split("/")
         val errorMessageBuilder = new StringBuilder
-        val effectiveRelationType = graphQLRelationOption.map(_.toAnnotation.relationType)
-        val newField = effectiveRelationType match {
-          case None | Some(FINDER) | Some(MULTI_GET) =>
+        val graphQLRelationTypeOption = graphQLRelationOption.map(_.toAnnotation.relationType)
+        val newField = graphQLRelationTypeOption match {
+          case None => // include-only relation
+            val newField = new RecordDataSchema.Field(new NullDataSchema)
+            newField.setOptional(true)
+            newField
+          case Some(FINDER) | Some(MULTI_GET) =>
             val newField = new RecordDataSchema.Field(new ArrayDataSchema(new StringDataSchema))
             newField.setOptional(false)
             newField
@@ -116,12 +121,12 @@ object Types extends StrictLogging {
           case Some(unknown) =>
             throw new RuntimeException(s"Unknown relation type: ${unknown.toString}")
         }
-        val graphQLProperty =
+        val graphQLPropertyOption =
           graphQLRelationOption.map(Relations.RELATION_PROPERTY_NAME -> _.toAnnotation.data)
-        val includeProperty =
-          relationOption.map(Relations.includedPropertyName -> _.toAnnotation.data)
+        val includePropertyOption =
+          relationOption.map(Relations.INCLUDED_PROPERTY_NAME -> _.toAnnotation.data)
         newField.setProperties(
-          List(graphQLProperty, includeProperty).flatten.toMap[String, AnyRef].asJava)
+          List(graphQLPropertyOption, includePropertyOption).flatten.toMap[String, AnyRef].asJava)
         graphQLRelationOption.map(_.description).foreach(newField.setDoc)
         newField.setName(nameSegments.last, errorMessageBuilder)
         insertFieldAtLocation(mergedSchema, nameSegments.dropRight(1).toList, newField)
--- a/naptime/src/main/scala/org/coursera/naptime/Types.scala
+++ b/naptime/src/main/scala/org/coursera/naptime/Types.scala
@@ -95,7 +95,7 @@ object Types extends StrictLogging {
     val relationKeys = fields.relations.keys ++ fields.graphQLRelations.keys
     for (name <- relationKeys) {
       val relationOption = fields.relations.get(name)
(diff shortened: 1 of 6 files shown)


naptime/src/main/scala/org/coursera/naptime/Types.scala
```diff
@@ -40,6 +40,7 @@ object Types extends StrictLogging {
 
   object Relations {
     val RELATION_PROPERTY_NAME = "relatedOn"
+    val includedPropertyName = "included"
```
cliu587, line 43: nit: for consistency, let's do `SHOUTY_CASE` for this variable as well.

naptime/src/test/scala/org/coursera/naptime/TypesTest.scala
```diff
@@ -91,4 +96,44 @@ class TypesTest extends AssertionsForJUnit {
...
+    assert(includeOnlyField.getType == new ArrayDataSchema(new StringDataSchema))
+    assert(includeOnlyField.getProperties.asScala === includeOnlyExpectedProperties)
+    val sharedField = resultingType.getField("shared")
+    val sharedExpectedProperties = Map(
+      "included" -> ResourceName("shared", 2).toAnnotation.data,
+      "relatedOn" -> GetGraphQLRelation(ResourceName("shared", 2), "get").toAnnotation.data
+    )
+    assert(sharedField != null)
+    assert(sharedField.getOptional)
+    assert(sharedField.getType == new StringDataSchema)
```
cliu587, line 129: This seems a bit confusing since it's different from L120 but it's because we're using a Get relation, maybe we can use the same GraphQL relation type in both cases so there's not this inconsistency?

naptime/src/main/scala/org/coursera/naptime/Types.scala
```diff
@@ -91,34 +92,39 @@ object Types extends StrictLogging {
...
-                "unknown relation type: " +
-                  graphQLRelation.toAnnotation.relationType.toString)
-          }
-          val relationMap = Map[String, AnyRef](
-            Relations.RELATION_PROPERTY_NAME -> graphQLRelation.toAnnotation.data())
-          newField.setProperties(relationMap.asJava)
-          newField.setDoc(graphQLRelation.description)
-          newField.setName(name.split("/").last, errorMessageBuilder)
-          insertFieldAtLocation(mergedSchema, name.split("/").dropRight(1).toList, newField)
+    val relationKeys = fields.relations.keys ++ fields.graphQLRelations.keys
```
cliu587, line 95: Looks like we've refactored the logic around a bit, the old logic seems really hard to read, so I'm having a hard time ensuring the new logic is equivalent, although I don't see any issues just by staring at a comparison between the two a few times. Maybe for testing, we can build a service with this artifact and hit the schema endpoint and do a comparison?

naptime/src/main/scala/org/coursera/naptime/Types.scala
```diff
@@ -91,34 +92,39 @@ object Types extends StrictLogging {
...
+      if (mergedSchema.contains(name)) {
+        logger.warn(
+          s"Fields for resource $typeName tries to add Include/GraphQL relation on field " +
+            s"'$name', but that field is already defined on the model")
+      } else {
+        val nameSegments = name.split("/")
+        val errorMessageBuilder = new StringBuilder
+        val effectiveRelationType = graphQLRelationOption.map(_.toAnnotation.relationType)
+        val newField = effectiveRelationType match {
+          case None | Some(FINDER) | Some(MULTI_GET) =>
```
cliu587, line 108: I think the `None` case should be called out explicitly, from my reading, in that case, relatedOption must not be none.

In that case, is `new RecordDataSchema.Field(new ArrayDataSchema(new StringDataSchema))` on L109 always correct for the related resource?

naptime/src/main/scala/org/coursera/naptime/Types.scala
cliu587, line 95: nit: should this be distincted? it could be that they are already both sets so this is not a problem but it's not clear just reading this code.

naptime/src/main/scala/org/coursera/naptime/Types.scala
```diff
@@ -91,34 +92,39 @@ object Types extends StrictLogging {
...
-        case Some(field) =>
-          logger.warn(
-            s"Fields for resource $typeName tries to add graphql relation on field " +
-              s"'$name', but that field is already defined on the model")
-        case None =>
-          val errorMessageBuilder = new StringBuilder
-          val newField = graphQLRelation.toAnnotation.relationType match {
-            case FINDER | MULTI_GET =>
-              val newField = new RecordDataSchema.Field(new ArrayDataSchema(new StringDataSchema)) // TODO(bryan): fix type here
-              newField.setOptional(false)
```
cliu587, line 105: Fishing for an answer here: do we know why we set optional: false for finder/multigets (and now related resources) while it's true for single element finder /gets?

cliu587: We resolved in DMs, looks like it's because finder/multigets return arrays so we can always fall back to empty list. This is the case for related resources as well.

naptime/src/main/scala/org/coursera/naptime/Types.scala
```diff
@@ -91,34 +92,39 @@ object Types extends StrictLogging {
...
+          case Some(SINGLE_ELEMENT_FINDER) | Some(GET) =>
+            val newField = new RecordDataSchema.Field(new StringDataSchema)
+            newField.setOptional(true)
+            newField
+          case Some(unknown) =>
+            throw new RuntimeException(s"Unknown relation type: ${unknown.toString}")
+        }
+        val graphQLProperty =
+          graphQLRelationOption.map(Relations.RELATION_PROPERTY_NAME -> _.toAnnotation.data)
+        val includeProperty =
```
cliu587, line 121: super nit: I think suffixing these with `Option` helps a lot with readability when not using an IDE here since we're flattening them on L124.

naptime/src/test/scala/org/coursera/naptime/TypesTest.scala
ppaskaris-coursera: I can, but I did it this way because I wanted to exercise both branches of the `match` with `Some`. I can also add a comment explaining the difference.

naptime/src/main/scala/org/coursera/naptime/Types.scala
ppaskaris-coursera: I had difficulty reasoning about it too, which is why I wrote the overlapping test case actually! I was going to try exercising it in `testInProd` with a `publishLocal` for an enterprise service, and fetch the schema with `curl` from the instance.

naptime/src/main/scala/org/coursera/naptime/Types.scala
ppaskaris-coursera: Yeah, the `relatedOption != None` is implied since we iterate over the keySet. Actually, now that I think about it more, I think this executes multiple times because I didn't union the keys, I just concatenated their iterables.

I'm going to go back on my idea here and give it a specific `null` type instead. I don't plan on using it in my type generator, and it serves no value to Assembler being the same. Plus, if the relation is 1-1, then when they add a GraphQL relation, it'll change types anyway, I'm not sure why I thought I could protect against that.

naptime/src/main/scala/org/coursera/naptime/Types.scala
ppaskaris-coursera: Yes, I just noticed this as well -- it should be distincted.

ppaskaris-coursera:
Whoops, very first step in publishing is to bump the version number, which I should have done here. I'll integration test it with publishLocal first and then come back with a version bump PR to roll it out.

cliu587:
Sounds good. You can bump the version and publish locally, I've had some issues with artifact caching when publishing a version that's already existent.

PR #282, opened 2020-12-17
Omit query analysis error message

gargrishabh75:
JIRA Ticket- https://coursera.atlassian.net/browse/FCORE-562

As part of the security ticket created we want to omit extra information provided by Assembler in case of QueryAnalysisError.

### DIFF
--- a/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/GraphQLController.scala
+++ b/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/GraphQLController.scala
@@ -147,7 +147,11 @@ class GraphQLController @Inject()(
               }
           }.recover {
             case error: QueryAnalysisError =>
-              OutgoingQuery(Json.obj("error" -> "QueryAnalysisError", "message" -> "message omitted for security"),None)
+              OutgoingQuery(
+                Json.obj(
+                  "error" -> "QueryAnalysisError",
+                  "message" -> "message omitted for security"),
+                None)
             case error: ErrorWithResolver =>
               OutgoingQuery(error.resolveError.as[JsObject], None)
             case error: Exception =>
--- a/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/GraphQLController.scala
+++ b/naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/GraphQLController.scala
@@ -16,6 +16,7 @@
 
 package org.coursera.naptime.ari.graphql.controllers
 
+import java.util.UUID.randomUUID
 import javax.inject._
 import com.typesafe.scalalogging.Logger
 import com.typesafe.scalalogging.StrictLogging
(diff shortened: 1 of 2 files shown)


naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/GraphQLController.scala
```diff
@@ -147,7 +147,11 @@ class GraphQLController @Inject()(
               }
           }.recover {
             case error: QueryAnalysisError =>
-              OutgoingQuery(error.resolveError.as[JsObject], None)
+              OutgoingQuery(
+                Json.obj(
```
cliu587, line 151: Do we need to `logger.warn` or `logger.error` here? is this frequent or useful?

cliu587 (approved):

naptime-graphql/src/main/scala/org/coursera/naptime/ari/graphql/controllers/GraphQLController.scala
gargrishabh75: This is a validation error, should we still log this?

kmcmurtrieca: Is there still a means to get the error?  Usually this is handled by logging the full error with unique token and then returning just that token through the API.

gargrishabh75: This is just a temporary solution that we wanted to release before the code freeze tomorrow. We will work on a proper fix later.

kmcmurtrieca: It doesn't have to be fancy.  Create a random UUID, log it and the error, then return the UUID.  I don't think it's a good idea to obfuscate diagnostics right before the holidays.

ricky-wong: +1 for `errorCode`, and I feel like `message omitted for security` is undesired.

kmcmurtrieca: There are other error handlers returning text: 'case error: Exception', 'case error: ErrorWithResolver ', and GraphQLController.exceptionHandler(...).  That last one has a TODO that looks security related.

ricky-wong: Considering that this has been long-existing behavior, it's more fine to keep it through the holidays than to jam in a potentially-disruptive change just because of the code freeze.

gargrishabh75: I am working on generating random UUID and logging it as @kmcmurtrieca suggested

kmcmurtrieca (approved):
This change looks good, but may be incomplete based on the remaining TODO in the original code.

PR #284, opened 2022-03-18
Serialize Courier default values

elin-coursera2:
Courier default values were not correctly being serialized to JSON because the previous `Writes` logic iterated through fields contained in the underlying `DataMap`, which would not include unset fields with default values. This PR addresses this issue by iterating through the fields declared in the record schema instead, and explicitly uses the default value if a field is unset and a default exists.


jeremycarroll (approved):

project/plugins.sbt
```diff
@@ -6,8 +6,6 @@ addSbtPlugin("org.coursera.courier" % "courier-sbt-plugin" % "2.1.4")
 
 addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "1.1")
 
-addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.0.0")
```
elin-coursera2, line 9: I removed this because this dependency is so old that it was removed from Maven. There's a newer version at `com.github.sbt:sbt-pgp:2.1.2`, but it doesn't seem to be compatible with sbt 0.13.

I *think* this still works because it seems sbt 0.13 actually bundles a copy of the original `com.jsuereth` jar.

elin-coursera2:
This didn't actually seem to fix the issue when I tested it

